PR #5, opened 2017-08-20
Documentation

gronke:
This pull-request summarizes the documentation of public and private API classes/methods.

### DIFF
--- a/libiocage/lib/Jail.py
+++ b/libiocage/lib/Jail.py
@@ -244,7 +244,7 @@ class Jail:
 
         return successful
 
-    def create(self, release_name, auto_download=False):
+    def create(self, release_name):
         """
         Create a Jail from a Release
 
@@ -252,11 +252,6 @@ class Jail:
 
             release_name (string):
                 The jail is created from the release matching the name provided
-
-            auto_download (bool): (default=False)
-                When no local release was found, but one with the matching name
-                is available on remote servers, the release is automatically
-                fetched.
         """
 
         self.require_jail_not_existing()
--- a/libiocage/lib/Jail.py
+++ b/libiocage/lib/Jail.py
@@ -18,8 +18,67 @@ import os
 
 
 class Jail:
+    """
+    Iocage unit orchestrates a jail's configuration and manages state
+
+    Jails are represented as zfs dataset `zpool/iocage/jails/<NAME>`
+
+    Directory Structure:
+
+        zpool/iocage/jails/<NAME>: The jail's dataset containing it's
+            configuration and root directory. Iocage legacy used to store
+            a jails configuration as ZFS properties on this dataset. Even
+            though the modern JSON config mechanism is preferred.
+
+        zpool/iocage/jails/<NAME>/root: This directory is the dataset
+            used as jail path when starting a jail. Usually the origin
+            or a root dataset is a release root directory
+
+        zpool/iocage/jails/<NAME>/config.json: Jails configured with the latest
+            configuration style store their information in a JSON file. When
+            this file is found in the jail's dataset, libiocage assumes
+            the jail to be a JSON-style jail and ignores other configuration
+            mechanisms
+
+        zpool/iocage/jails/<NAME>/config: Another supported configuration
+            mechanism is a UCL file. It's content is only taken into account if
+            no JSON or ZFS configuration was found
+
+    Jail Types:
+
+        Standalone: The /root dataset gets cloned from a release at creation
+            time. It it not affected by changes to the Release and persists all
+            data within the jail
+
+        NullFS Basejail: The fastest method to spawn a basejail by mounting
+            read-only directories from the release's root dataset by creating
+            a snapshot of the release on each boot of the jail. When a release
+            was updated, the jail is updated as well on the next reboot. This
+            type is the one used by the Python implementation of iocage.
+
+        ZFS Basejail: Legacy basejails used to clone individual datasets from a
+            release (stored in `zpool/iocage/base/<RELEASE>)
+    """
 
     def __init__(self, data={}, zfs=None, host=None, logger=None, new=False):
+        """
+        Initializes a Jail
+
+        Args:
+
+            data (string|dict):
+                Jail configuration dict or jail name as string identifier.
+
+            zfs (libzfs.ZFS): (optional)
+                Inherit an existing libzfs.ZFS() instance from ancestor classes
+
+            host (libiocage.lib.Host): (optional)
+                Inherit an existing Host instance from ancestor classes
+
+            logger (libiocage.lib.Logger): (optional)
+                Inherit an existing Logger instance from ancestor classes
+
+        """
 
         libiocage.lib.helpers.init_logger(self, logger)
         libiocage.lib.helpers.init_zfs(self, zfs)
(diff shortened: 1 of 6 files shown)


libiocage/lib/Jail.py
```diff
@@ -18,6 +18,49 @@
 
 
 class Jail:
+    """
+    Iocage unit orchestrates a jail's configuration and manages state
```
skarekrow: Perhaps say `iocage Jail unit`? Remember `iocage` is always lowercase, it's our name. That goes for all over.

libiocage/lib/Jail.py
```diff
@@ -18,6 +18,49 @@
...
+
+    Jails are represented as zfs dataset `zpool/iocage/jails/<NAME>`
```
skarekrow: `as a zfs dataset`

libiocage/lib/Jail.py
```diff
@@ -18,6 +18,49 @@
...
+
+    Directory Structure:
+
+        zpool/iocage/jails/<NAME>: The jail's dataset containing it's
+            configuration and root directory. Iocage legacy used to store
```
skarekrow: lowercase `iocage`, we can drop and root directory. It's not a directory, but a dataset. And isn't terribly important in this description. I think we should drop the JSON being preferred, and make the language stronger. It's the only supported mechanism. There are a lot of new things and properties, and this will help ease the migrations to it if we only support that mechanism.

libiocage/lib/Jail.py
```diff
@@ -18,6 +18,49 @@
...
+            a jails configuration as ZFS properties on this dataset. Even
+            though the modern JSON config mechanism is preferred.
+
+        zpool/iocage/jails/<NAME>/root: This directory is the dataset
```
skarekrow: This can just be said as `This dataset is used as the jail's root when starting the jail`. Perhaps instead of origin we say `a clone's root or release's root dataset are normally the sources`

libiocage/lib/Jail.py
```diff
@@ -18,6 +18,49 @@
...
+            used as jail path when starting a jail. Usually the origin
+            or a root dataset is a release root directory
+
+        zpool/iocage/jails/<NAME>/config.json: Jails configured with the latest
+            configuration style store their information in a JSON file. When
+            this file is found in the jail's dataset, libiocage assumes
+            the jail to be a JSON-style jail and ignores other configuration
+            mechanisms
+
+        zpool/iocage/jails/<NAME>/config: Another supported configuration
```
skarekrow: Drop the word supported, it isn't supported, but is able to be read. We need to offer a cutoff, we don't want to juggle supporting 3 different configuration methods.

libiocage/lib/Jail.py
```diff
@@ -18,6 +18,49 @@
...
+    Jail Types:
+
+        Standalone: The /root dataset gets cloned from a release at creation
+            time. It it not affected by changes to the Release and persists all
+            data within the jail
+
+        NullFS Basejail: The fastest method to spawn a basejail by mounting
+            read-only directories from the release's root dataset by creating
+            a snapshot of the release on each boot of the jail. When a release
+            was updated, the jail is updated as well on the next reboot. This
```
skarekrow: `When a release is updated` not was. We should clarify that the jails userland will be updated, but not the configuration in `/etc`. The user would have to check that.

libiocage/lib/Jail.py
```diff
@@ -153,6 +245,20 @@ def force_stop(self):
...
     def create(self, release_name, auto_download=False):
+        """
+        Create a Jail from a Release
+
+        Args:
+
+            release_name (string):
+                The jail is created from the release matching the name provided
+
+            auto_download (bool): (default=False)
```
igalic: `auto_download` seems to be unused.

libiocage/lib/Jail.py
```diff
@@ -415,34 +546,49 @@ def _set_route(self, gateway, ipv6=False):
...
                 logger=self.logger
             )
 
     def update_jail_state(self):
+        """
+        Invoke update of the jail state from jls output
+        """
         try:
             stdout = subprocess.check_output([
                 "/usr/sbin/jls",
```
igalic: consiser using `--libxo=json` here, for easier (no) parsing.

libiocage/lib/Jail.py
```diff
@@ -499,10 +646,21 @@ def _resolve_name(self, text):
 
         raise libiocage.lib.errors.JailNotFound(text, logger=self.logger)
 
-    def _get_name(self):
+    @property
+    def name(self):
+        """
+        The name (formerly UUID) of the Jail
+        """
         return self.config["id"]
```
igalic: how does the setter work here?
(i.e.: how do we add validation for this?)

libiocage/lib/Jail.py
```diff
@@ -601,7 +797,14 @@ def __getattr__(self, key):
 
         raise AttributeError(f"Jail property {key} not found")
 
-    def getattr_str(self, key):
+    def getstring(self, key):
```
igalic: why is this a public function rather than a `__str__` on the properties?
or would that be too convoluted?

libiocage/lib/helpers.py
```diff
@@ -72,7 +72,7 @@ def exec(command, logger=None, ignore_error=False):
     if child.returncode > 0:
 
         if logger:
-            log_level = "spam" if ignore_error else "warning"
+            log_level = "spam" if ignore_error else "warn"
```
igalic: starting to think our logger levels should also be tuples.

libiocage/lib/Jail.py
gronke: Good catch, haha! This can be implemented on CLI side, so no need to expose an interface to automatically fetch the release here. Thanks!

libiocage/lib/Jail.py
gronke: 💛! But this will go into a separate PR.

libiocage/lib/Jail.py
gronke: This is write-only here. Just an abstraction of a jails given identifier. I'm sure we can improve validation in JailConfig.py

libiocage/lib/Jail.py
gronke: That would require to wrap all types returned to return a `-` in case the original value is `None`.

libiocage/lib/JailConfig.py
```diff
@@ -63,25 +63,28 @@ def read(self):
             libiocage.lib.JailConfigJSON.JailConfigJSON.read(self)
             self["legacy"] = False
             self.logger.log("Configuration loaded from JSON", level="verbose")
-            return
+            return "json"
```
gronke: We should introduce tuples in another PR. Good idea.

libiocage/lib/Jail.py
igalic: i wonder if that's an improvement in any regard, esp for downstream users who consume our output automatically.

gronke: No it's not. When you don't intent to print the outputs, you don't want to get the strings. Some objects, like for example the JailConfigAddresses, are a custom data structure you want to have access to.

PR #6, opened 2017-08-21
fix our installation (instructions)

igalic:
* chore(build): install cython
* doc(build): most systems will need libgit2 libucl

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,7 +3,7 @@ SERVER=""
 
 install:
 	python3.6 -m ensurepip
-	pkg install -q -y libgit2
+	pkg install -q -y libgit2 libucl cython3
 	pip3.6 install -U .
 uninstall:
 	pip3.6 uninstall -y libiocage
--- a/README.md
+++ b/README.md
@@ -31,6 +31,8 @@ cd libiocage
 make install
 ```
 
+Please note: this will build `py-libzfs` from source, which will require `/usr/src` to be populated.
+
 ## Usage
 
 ### Library


skarekrow:
While it LGTM, I do not really like that we have a hard dependency on libucl. As many users may not be coming from `iocage-legacy`'s `develop` branch, they wouldn't need it. It would be nice to make this optional.

Makefile
```diff
@@ -3,7 +3,7 @@ SERVER=""
 
 install:
 	python3.6 -m ensurepip
-	pkg install -q -y libgit2
+	pkg install -q -y libgit2 cython
```
igalic: this should be `pip3.6 install cython`

README.md
```diff
@@ -26,6 +26,7 @@ pip3.6 install libiocage
 ### from Master branch
 
 ```
+pkg install -y libgit2 libucl
```
igalic: this line shouldn't be necessary at all…

Makefile
igalic: instead, we need libucl here.

igalic: we *could* install cython3 however, from pkg, it would make things a lot less painful.

skarekrow:
Still LGTM

Makefile
```diff
@@ -3,7 +3,7 @@ SERVER=""
 
 install:
 	python3.6 -m ensurepip
-	pkg install -q -y libgit2
+	pkg install -q -y libgit2 libucl cython3
```
igalic: this works, but only if we don't use our own venv…
which, per our docs, we don't (recommend)

gronke:
@skarekrow the hard dependency on libucl is only for developers. We can argue if we want to introduce configurations to enable or disable features, but it seems more acceptable to bundle libucl than making a special case out of it. Ah, btw. libucl is also used to parse rc.conf files 😏

skarekrow:
@gronke fair enough. ;)

PR #17, opened 2017-08-22
validate names, conservatively: ascii, 32 char

### LINKED ISSUE #10: validate names helper
almost all of our `name` arguments have the same general format:

```perl
/^[a-z0-9][a-z0-9-_\.]*/i
```

given that many of our names are used as part of a path, we can further restrict the `*`

```perl
/^[a-z0-9][a-z0-9-_\.]{0,254}/i
```

to FreeBSD's `NAME_MAX` (255).

****
> edit with conclusions from the comments

HOWEVER, until FreeBSD 12, `fstab`'s mount paths are further restricted to 88, and not 255. Since we're supporting freebsd all the way down to 9.3, we should also limit our names.

As we've seen from experience from using UUIDs as names, 32 is a length that can stretch it. I suggest only using half of that (16)

```perl
/^[a-z0-9][a-z0-9-_\.]{0,15}/i
```

igalic:
This fixes #10.

### DIFF
--- a/libiocage/lib/Datasets.py
+++ b/libiocage/lib/Datasets.py
@@ -135,6 +135,9 @@ class Datasets:
                                pool=None,
                                mountpoint=None):
 
+        if not libiocage.lib.helpers.validate_name(name):
+            raise NameError(f"Invalid 'name' for Dataset: {name}")
+
         try:
             return self.datasets[name]
         except (AttributeError, KeyError):
--- a/libiocage/lib/Release.py
+++ b/libiocage/lib/Release.py
@@ -36,6 +36,9 @@ class Release:
         libiocage.lib.helpers.init_zfs(self, zfs)
         libiocage.lib.helpers.init_host(self, host)
 
+        if not libiocage.lib.helpers.validate_name(name):
+            raise NameError(f"Invalid 'name' for Release: '{name}'")
+
         self.name = name
         self._hashes = None
         self._dataset = None
--- a/libiocage/lib/helpers.py
+++ b/libiocage/lib/helpers.py
@@ -1,4 +1,5 @@
 import subprocess
+import re
 
 import libzfs
 
@@ -93,6 +94,12 @@ def _prettify_output(output):
         lambda line: f"    {line}",
         output.split("\n")
     ))
+
+
+# helper function to validate names
+def validate_name(name):
+    validate = re.compile(r'[a-z0-9][a-z0-9\.-_]{0,31}', re.I)
+    return bool(validate.fullmatch(name))
 
 
 def exec_passthru(command, logger=None):


libiocage/lib/Datasets.py
```diff
@@ -135,6 +135,9 @@ def _get_or_create_dataset(self,
                                pool=None,
                                mountpoint=None):
 
+        if not libiocage.lib.helpers.validate_name(name):
+            raise NameError(f"Unable to validate 'name' ({name}) for Dataset.")
```
igalic: these should be consistent with the other message

libiocage/lib/helpers.py
```diff
@@ -95,6 +96,13 @@ def _prettify_output(output):
     ))
 
 
+# helper function to validate names
+def validate_name(name):
+    validate = re.compile(r'^[a-z0-9][a-z0-9\.-_]{0,31}$', re.I)
+    validate.fullmatch(name)
+    return True if validate.fullmatch(name) else False
```
gronke: Can be reduced to `bool(validate.fullmatch(name))`

libiocage/lib/helpers.py
```diff
@@ -95,6 +96,13 @@ def _prettify_output(output):
     ))
 
 
+# helper function to validate names
+def validate_name(name):
+    validate = re.compile(r'^[a-z0-9][a-z0-9\.-_]{0,31}$', re.I)
```
gronke: You do a .fullmatch later on, so you can omit `^` and `$`

igalic:
thanks for your review, @gronke.
I'll fix this up tomorrow on the train

libiocage/lib/Release.py
```diff
@@ -36,6 +36,9 @@ def __init__(self, name=None,
         libiocage.lib.helpers.init_zfs(self, zfs)
         libiocage.lib.helpers.init_host(self, host)
 
+        if not libiocage.lib.helpers.validate_name(name):
+            raise NameError(f"Invalid 'name' for Release: '{name}'")
```
skarekrow: Why are we validating RELEASE? They aren't created by a user with a name, they come from the project. We should remove this.

gronke: Releases are stored with the same mechanism as Jails are, so that the same limitations apply.

gronke:
@igalic ready to merge! Can you please rebase?

libiocage/lib/Release.py
skarekrow: I wasn't thinking right, @gronke is correct. Free to merge @igalic

igalic: validate all the things, imo

PR #20, opened 2017-08-24
consistently check & set rcconf booleans

igalic:
***do not merge***
until i have written tests ;)


gronke:
@igalic I consider this a bugfix. We need to translate between strings like "Yes", "YES", "1", "true", etc all the time and should have a parser in one place throughout the library. I will create a ticket to refactor this!

PR #22, opened 2017-08-24
add helper for parsing boolean values

igalic:
this helper function will now consistently read all manner of strings
and return the appropriate boolean value (or the passed default)

****

in the next step, we will need a write-helper for translating in the
other direction.

### DIFF
--- a/libiocage/lib/JailConfig.py
+++ b/libiocage/lib/JailConfig.py
@@ -185,21 +185,22 @@ class JailConfig(dict, object):
         return False
 
     def _set_basejail(self, value, **kwargs):
-        enabled = libiocage.lib.helpers.parse_bool(value)
         if self["legacy"]:
-            self.data["basejail"] = "on" if enabled else "off"
+            self.data["basejail"] = libiocage.lib.helpers.get_str_bool(
+                value, true="on", false="off")
         else:
-            self.data["basejail"] = "yes" if enabled else "no"
+            self.data["basejail"] = libiocage.lib.helpers.get_str_bool(
+                value, true="yes", false="no")
 
     def _get_clonejail(self):
-        return self.data["clonejail"] == "on"
+        return libiocage.lib.helpers.parse_bool(self.data["clonejail"])
 
     def _default_clonejail(self):
         return True
 
     def _set_clonejail(self, value, **kwargs):
-        self.data["clonejail"] = "on" if (value is True) or (
-                                         value == "on") else "off"
+        self.data["clonejail"] = libiocage.lib.helpers.get_str_bool(
+            value, true="on", false="off")
 
     def _get_ip4_addr(self):
         try:
@@ -273,11 +274,11 @@ class JailConfig(dict, object):
         return None
 
     def _get_vnet(self):
-        return self.data["vnet"] == "on"
+        return libiocage.lib.helpers.parse_bool(self.data["vnet"])
 
     def _set_vnet(self, value, **kwargs):
-        vnet_enabled = (value == "on") or (value is True)
-        self.data["vnet"] = "on" if vnet_enabled else "off"
+        self.data["vnet"] = libiocage.lib.helpers.get_str_bool(
+            value, true="on", false="off")
 
     def _get_jail_zfs_dataset(self):
         try:
@@ -291,20 +292,19 @@ class JailConfig(dict, object):
         self.data["jail_zfs_dataset"] = " ".join(value)
 
     def _get_jail_zfs(self):
-        enabled = self.data["jail_zfs"] == "on"
+        enabled = libiocage.lib.helpers.parse_bool(self.data["jail_zfs"])
         if not enabled:
             if len(self.jail_zfs_dataset) > 0:
                 raise libiocage.lib.errors.JailConigZFSIsNotAllowed(
-                    logger=self.logger
-                )
+                    logger=self.logger)
         return enabled
 
     def _set_jail_zfs(self, value, **kwargs):
         if (value is None) or (value == ""):
             del self.data["jail_zfs"]
             return
-        enabled = (value == "on") or (value is True)
-        self.data["jail_zfs"] = "on" if enabled else "off"
+        self.data["jail_zfs"] = libiocage.lib.helpers.get_str_bool(
+            value, true="on", false="off")
 
     def _default_jail_zfs(self):
         # if self.data["jail_zfs"] does not explicitly exist, _get_jail_zfs
--- a/libiocage/lib/JailConfig.py
+++ b/libiocage/lib/JailConfig.py
@@ -179,15 +179,13 @@ class JailConfig(dict, object):
             self.data["type"] = value
 
     def _get_basejail(self):
-        value = self.data["basejail"]
-        return (value == "on") or (value == "yes")
+        return libiocage.lib.helpers.parse_bool(self.data["basejail"])
 
     def _default_basejail(self):
         return False
 
     def _set_basejail(self, value, **kwargs):
-        # ToDo: generalize boolean value inputs
-        enabled = (value is True) or (value == "on") or (value == "yes")
+        enabled = libiocage.lib.helpers.parse_bool(value)
         if self["legacy"]:
             self.data["basejail"] = "on" if enabled else "off"
         else:
--- a/libiocage/lib/helpers.py
+++ b/libiocage/lib/helpers.py
@@ -116,14 +116,28 @@ def parse_bool(data, default=False):
     True
     >>> parse_bool("false")
     False
-    >>> parse_bool("-")
+    >>> parse_bool("/opt/android-studio/build.txt")
     False
+
+    Note that "-" gets a special treatment:
+
+    >>> parse_bool("-")
+    None
+
+    The behavior of the default parameter can be used to create a
+    pass-thru function:
+
+    >>> data = "/opt/android-studio/build.txt"
+    >>> parse_bool(data, data)
+    "/opt/android-studio/build.txt"
     """
 
     if isinstance(data, bool):
         return data
     if isinstance(data, str):
-        if data.lower() in ["yes", "true", "on", "1"]:
+        if data == "-":
+            return None
+        elif data.lower() in ["yes", "true", "on", "1"]:
             return True
         elif data.lower() in ["no", "false", "off", "0"]:
             return False
--- a/libiocage/lib/helpers.py
+++ b/libiocage/lib/helpers.py
@@ -144,9 +144,10 @@ def parse_bool(data, default=False):
     return default
 
 
-def try_parse_bool(data):
+def parse_user_input(data):
     """
-    like parse_bool(), but returns the input itself if parsing fails
+    uses parse_bool() to partially return Boolean and NoneType values
+    All other types as returned as-is
 
     >>> parse_bool("YES")
     True
--- a/libiocage/lib/helpers.py
+++ b/libiocage/lib/helpers.py
@@ -103,6 +103,47 @@ def validate_name(name):
     return bool(validate.fullmatch(name))
 
 
+# data -> bool | default
+def parse_bool(data, default=False):
+    """
+    try to parse booleans from strings
+
+    On success, it returns the parsed boolean
+    on failure it returns the `default`.
+    By default, `default` is `False`.
+
+    >>> parse_bool("YES")
+    True
+    >>> parse_bool("false")
+    False
+    >>> parse_bool("-")
+    False
+    """
+
+    if isinstance(data, bool):
+        return data
+    if isinstance(data, str):
+        if data.lower() in ["yes", "true", "on", "1"]:
+            return True
+        elif data.lower() in ["no", "false", "off", "0"]:
+            return False
+    return default
+
+
+def try_parse_bool(data):
+    """
+    like parse_bool(), but returns the input itself if parsing fails
+
+    >>> parse_bool("YES")
+    True
+    >>> parse_bool("false")
+    False
+    >>> parse_bool(8.4)
+    8.4
+    """
+    return parse_bool(data, data)
+
+
 def exec_passthru(command, logger=None):
     if isinstance(command, str):
         command = [command]
--- a/libiocage/lib/helpers.py
+++ b/libiocage/lib/helpers.py
@@ -144,6 +144,23 @@ def try_parse_bool(data):
     return parse_bool(data, data)
 
 
+def get_str_bool(data, true="yes", false="no"):
+    """
+    return a string boolean value using parse_bool(), of specified style
+
+    >>> get_str_bool(True)
+    "yes"
+    >>> get_str_bool(False)
+    "no"
+
+    >>> get_str_bool(True, true="yip", false="nope")
+    "yip"
+    >>> get_str_bool(False, true="yip", false="nope")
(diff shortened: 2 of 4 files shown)


libiocage/lib/JailConfig.py
```diff
@@ -179,15 +179,13 @@ def _set_type(self, value, **kwargs):
...
-        return (value == "on") or (value == "yes")
+        return libiocage.lib.helpers.parse_bool(self.data["basejail"])
 
     def _default_basejail(self):
         return False
 
     def _set_basejail(self, value, **kwargs):
-        # ToDo: generalize boolean value inputs
-        enabled = (value is True) or (value == "on") or (value == "yes")
+        enabled = libiocage.lib.helpers.parse_bool(value)
```
igalic: oops, i didn't notice the `(value is True)` bit

libiocage/lib/helpers.py
```diff
@@ -102,6 +102,26 @@ def validate_name(name):
     return bool(validate.fullmatch(name))
 
 
+# data -> bool | default
+# This little helper function will try to parse booleans from strings
+# On success, it returns the parsed boolean, on failure it returns the
+# `default`. By default, `default` is `False`.
+def parse_bool(data, default=False):
```
gronke: According to pydoc you can put your docstrings in the next line below the function definition. Would you? :)

libiocage/lib/helpers.py
```diff
@@ -102,6 +102,26 @@ def validate_name(name):
...
+        return data
+    if isinstance(data, str):
+        if data.lower() in ["yes", "true", "on", "1"]:
+            return True
+        elif data.lower() in ["no", "false", "off", "0"]:
+            return False
+    return default
+
+
+# same as above, but returns the input if parsing fails
```
gronke: Same with this comment - it can become a pydoc string (maybe you want to add some more description).

libiocage/lib/helpers.py
```diff
@@ -102,6 +102,26 @@ def validate_name(name):
...
 
+# data -> bool | default
+# This little helper function will try to parse booleans from strings
+# On success, it returns the parsed boolean, on failure it returns the
+# `default`. By default, `default` is `False`.
+def parse_bool(data, default=False):
+    if isinstance(data, bool):
+        return data
+    if isinstance(data, str):
+        if data.lower() in ["yes", "true", "on", "1"]:
```
gronke: True or int(1) are valid values as well.

libiocage/lib/helpers.py
skarekrow: In addition it's odd to say it returns the default which is False. Why not just say else it returns False?

libiocage/lib/helpers.py
skarekrow: If `int` is valid, why are we checking for `str` only? It should be a separate `elif` imo in that case.

libiocage/lib/helpers.py
igalic: so, i was playing with this, because there were a bunch of different use-cases:
but most of the time, the comparison was directly against a `Truthy` value, which meant to me, that in absence of truth, false would've been the _default_
However, a number of other cases use it in such a away that the value is essentially passed-thru, but modified in case it can be interpreted as `Truthy` or `Falsey`.

libiocage/lib/helpers.py
igalic: `Int(1)` is tricky, because it can also simply be 1.

libiocage/lib/helpers.py
```diff
@@ -102,6 +102,45 @@ def validate_name(name):
     return bool(validate.fullmatch(name))
 
 
+# data -> bool | default
+def parse_bool(data, default=False):
+    """try to parse booleans from strings
```
skarekrow: Nitpick, let's do a newline after `"""`

libiocage/lib/helpers.py
```diff
@@ -102,6 +102,45 @@ def validate_name(name):
...
+    if isinstance(data, str):
+        if data.lower() in ["yes", "true", "on", "1"]:
+            return True
+        elif data.lower() in ["no", "false", "off", "0"]:
+            return False
+    return default
+
+
+def try_parse_bool(data):
+    """like parse_bool(), but returns the input itself if parsing fails
```
skarekrow: Also here

gronke:
> @igalic Can you rebase this on master instead of a merge commit?

The repo settings enforce rebase merges. @skarekrow you can check out a developer's branch locally, rebase and merge this into master locally. You will be the committer of the commits after that, but the author remains the same, so no credit is lost. GitHub should detect this and automatically close the according PR as merged.

libiocage/lib/helpers.py
```diff
@@ -103,6 +103,47 @@ def validate_name(name):
...
+    On success, it returns the parsed boolean
+    on failure it returns the `default`.
+    By default, `default` is `False`.
+
+    >>> parse_bool("YES")
+    True
+    >>> parse_bool("false")
+    False
+    >>> parse_bool("-")
+    False
```
gronke: `-` should resolve to `None`. We also require this behavior in try_parse_bool later on.

libiocage/lib/helpers.py
```diff
@@ -103,6 +103,47 @@ def validate_name(name):
...
+        return data
+    if isinstance(data, str):
+        if data.lower() in ["yes", "true", "on", "1"]:
+            return True
+        elif data.lower() in ["no", "false", "off", "0"]:
+            return False
+    return default
+
+
+def try_parse_bool(data):
```
gronke: What about `parse_user_input` as method name? I think this function does more than the name suggests.

libiocage/lib/helpers.py
igalic: good catch

libiocage/lib/helpers.py
igalic: that's a bit… generous ;)

libiocage/lib/helpers.py
```diff
@@ -103,6 +103,67 @@ def validate_name(name):
...
+    "NO"
+    >>> get_str_bool(True, "ini")
+    "on"
+    >>> get_str_bool(False, "ini")
+    "off"
+    >>> get_str_bool(True, "jailconfig")
+    "yes"
+    >>> get_str_bool(False, "jailconfig")
+    "no"
+    """
```
igalic: this is bad, and i don't know if it's the correct way to do.

libiocage/lib/helpers.py
```diff
@@ -103,6 +103,67 @@ def validate_name(name):
...
+    return "on" if parse_bool(data) else "off"
```
igalic: yes, i'm aware it's not implemented yet ;)

libiocage/lib/JailConfig.py
```diff
@@ -199,9 +197,8 @@ def _get_clonejail(self):
     def _default_clonejail(self):
         return True
 
-    def _set_clonejail(self, value, **kwargs):
-        self.data["clonejail"] = "on" if (value is True) or (
-                                         value == "on") else "off"
+    def _set_clonejail(self, val, **kwargs):
+        self.data["clonejail"] = libiocage.lib.helpers.get_str_bool(val, "jailconfig")
```
igalic: there's no coherent way to set this so that it won't overflow 79 char.

skarekrow: Sure there is, line break after `str_bool(`:
```
        self.data["clonejail"] = libiocage.lib.helpers.get_str_bool(
            val,"jailconfig")
```

gronke: 🙄 In addition to previous comment, I'm in favor of this style:
```
        self.data["clonejail"] = libiocage.lib.helpers.get_str_bool(
            val,
            "jailconfig"
        )
```

Please let us not randomize wrapping. And @igalic I got your point that an 80 character limit is randomly choosen. The two lines you deleted here before looked super ugly. That destroys all benefit 80 characters width can have, because my eyes refuse to read it.. :(

skarekrow: That's a fine style too :P Point was, 80 characters was not a restriction here.

libiocage/lib/helpers.py
skarekrow: Are you referring to `style` not being implemented yet?

libiocage/lib/JailConfig.py
igalic: fixed.

libiocage/lib/helpers.py
```diff
@@ -103,6 +103,82 @@ def validate_name(name):
...
+    True
+    >>> parse_bool("false")
+    False
+    >>> parse_bool(8.4)
+    8.4
+    """
+    return parse_bool(data, data)
+
+
+def get_str_bool(data, style=None):
```
gronke: Just an idea: We could also use two attributes `yes` and `no` containing the strings we expect for True or False values. like this:

```python
def get_str_bool(data, yes="yes", no="no"):
  ...

>>> get_str_bool(True, yes="yip", no="nope")
"yip"
>>> get_str_bool(False, yes="yip", no="nope")
"nope"
```

igalic: how about a tuple instead of two parameters?

igalic: worst case scenario of a `style` tuple:

```python
def get_str_bool(data, style=("yes", "no",))
    return style[0] if parse_bool(data) else style[1]

>>> get_str_bool(True, style=("1", "0",))
"1"
```

libiocage/lib/JailConfig.py
```diff
@@ -293,20 +292,19 @@ def _set_jail_zfs_dataset(self, value, **kwargs):
...
                 raise libiocage.lib.errors.JailConigZFSIsNotAllowed(
-                    logger=self.logger
-                )
+                    logger=self.logger)
         return enabled
 
     def _set_jail_zfs(self, value, **kwargs):
         if (value is None) or (value == ""):
             del self.data["jail_zfs"]
             return
```
igalic: there's a bunch of places in this file where we use `stringify()` and i'm not sure if we need to look into that, too

igalic:
i love that most of these helper functions are a line long, with 15 lines of documentation…

libiocage/lib/helpers.py
gronke: As long as nobody invents a third boolean state for iocage... Personally I prefer the two parameters, because it seems more intuitive to me.

gronke:
> what should be done with stringify()?

We should remove it. The helper will be good to replace all occasions of the stringify function.

igalic:
> We should remove it.

please do. while i go sleep :P

libiocage/lib/helpers.py
```diff
@@ -103,6 +103,79 @@ def validate_name(name):
...
+
+    >>> parse_bool("-")
+    None
+
+    The behavior of the default parameter can be used to create a
+    pass-thru function:
+
+    >>> data = "/opt/android-studio/build.txt"
+    >>> parse_bool(data, data)
+    "/opt/android-studio/build.txt"
```
gronke: We should choose some more generic sample string, because otherwise we could accidentally show up in search results for android-studio and confuse people. What about `/etc/passwd`?

Can you rename the variable `data` to `default`. That's a bit more intuitive to read.

libiocage/lib/helpers.py
```diff
@@ -103,6 +103,79 @@ def validate_name(name):
...
+    """
+
+    if isinstance(data, bool):
+        return data
+    if isinstance(data, str):
+        if data == "-":
+            return None
+        elif data.lower() in ["yes", "true", "on", "1"]:
```
gronke: We need to be careful with `1` and `0`, because from user input they may be interpreted as string and resolved to a boolean. I'd say those values from user input do not resolve to boolean values at all.

libiocage/lib/helpers.py
```diff
@@ -103,6 +103,79 @@ def validate_name(name):
...
+    On success, it returns the parsed boolean
+    on failure it returns the `default`.
+    By default, `default` is `False`.
+
+    >>> parse_bool("YES")
+    True
+    >>> parse_bool("false")
+    False
+    >>> parse_bool("/opt/android-studio/build.txt")
+    False
```
gronke: A method called parse_bool should only parse booleans or raise. If it can read a boolean-value from the mapping table, that's fine. `None` is okay too, but everything else should raise.

Instead of giving the `parse_bool` method a `default` argument, you could use a try-block in `parse_user_input` that returns the input data in case of failure.

libiocage/lib/helpers.py
igalic: done

libiocage/lib/helpers.py
igalic: yeah, this is extremely context sensitive, and these functions don't have that context.

gronke: True, but I can't remember where iocage ever allowed 0/1 as boolean values. Let's just drop those two.

igalic: here's a full listing of potential candidates:

```
 ~/s/t/libiocage  ╍╍ bools  ag -B1 '"[01]"'
libiocage/lib/JailConfig.py
416-    def _default_children_max(self):
417:        return "0"
419-    def _default_allow_set_hostname(self):
420:        return "1"
422-    def _default_allow_sysvipc(self):
423:        return "0"
425-    def _default_allow_raw_sockets(self):
426:        return "0"
428-    def _default_allow_chflags(self):
429:        return "0"
431-    def _default_allow_mount(self):
432:        return "0"
434-    def _default_allow_mount_devfs(self):
435:        return "0"
437-    def _default_allow_mount_nullfs(self):
438:        return "0"
440-    def _default_allow_mount_procfs(self):
441:        return "0"
443-    def _default_allow_mount_zfs(self):
444:        return "0"
446-    def _default_allow_mount_tmpfs(self):
447:        return "0"
449-    def _default_allow_quotas(self):
450:        return "0"
452-    def _default_allow_socket_af(self):
453:        return "0"
464-    def _default_exec_clean(self):
465:        return "1"
467-    def _default_exec_fib(self):
468:        return "0"
494-    def _default_mount_devfs(self):
495:        return "1"
497-    def _default_mount_fdescfs(self):
498:        return "1"

libiocage/lib/JailConfigFstab.py
103-            options="ro",
104:            dump="0",
105:            passnum="0",
144-                "options"    : "ro",
145:                "dump"       : "0",
146:                "passnum"    : "0",

libiocage/lib/Release.py
314-            "basejail"          : False,
315:            "allow_mount_nullfs": "1",
316-            "release"           : self.name,
317:            "securelevel"       : "0"

libiocage/lib/helpers.py
139-            return None
140:        elif data.lower() in ["yes", "true", "on", "1"]:
141-            return True
142:        elif data.lower() in ["no", "false", "off", "0"]:

libiocage/cli/create.py
51-@click.command(name="create", help="Create a jail.")
52:@click.option("--count", "-c", callback=validate_count, default="1",
 ~/s/t/libiocage  ╍╍ bools  
```

gronke: So what's our response? Removing `0` and `1` from the values that resolve to boolean?

gronke:
@igalic your changes have been merged! Thanks so much for pointing your finger at it and implementing the solution 💛

PR #24, opened 2017-08-24
Fix regex for RELEASE

skarekrow:
This would fail on valid RELEASE's

### DIFF
--- a/libiocage/lib/helpers.py
+++ b/libiocage/lib/helpers.py
@@ -98,7 +98,7 @@ def _prettify_output(output):
 
 # helper function to validate names
 def validate_name(name):
-    validate = re.compile(r'[a-z0-9][a-z0-9\.-_-]{0,31}', re.I)
+    validate = re.compile(r'[a-z0-9][a-z0-9\.\-_]{0,31}', re.I)
 
     return bool(validate.fullmatch(name))
 
--- a/libiocage/lib/helpers.py
+++ b/libiocage/lib/helpers.py
@@ -98,7 +98,8 @@ def _prettify_output(output):
 
 # helper function to validate names
 def validate_name(name):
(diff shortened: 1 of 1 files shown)


libiocage/lib/helpers.py
```diff
@@ -98,7 +98,8 @@ def _prettify_output(output):
 
 # helper function to validate names
 def validate_name(name):
-    validate = re.compile(r'[a-z0-9][a-z0-9\.-_]{0,31}', re.I)
+    validate = re.compile(r'[a-z0-9][a-z0-9\.-_-]{0,31}', re.I)
```
igalic: why are there two `-`s now?

skarekrow: Because without it, for example -- `11.1-RELEASE` would only match `11.1`, when we do intend to match `11.1-RELEASE` in its entirety.

igalic:
```
In [43]: validate = re.compile(r'[a-z0-9][a-z0-9\.-_-]{0,31}', re.I)

In [44]: release = "11.1-RELEASE"

In [45]: bool(validate.fullmatch(release))
Out[45]: True

```

igalic: so, i think we want this:
```
validate = re.compile(r'[a-z0-9][a-z0-9\.\-_]{0,31}', re.I)
```

before, the regex meant: from `.` to `_`
you changed it to mean: from `.` to `_`, `-`
what we want is: `.`, `_`, `-`.

skarekrow: We don't, that doesn't match the word `RELEASE` afterwords.

skarekrow: And that's because I didn't get enough caffeine, @igalic is right. Changing that now

PR #30, opened 2017-08-25
HardenedBSD Support

### LINKED ISSUE #27: HardenedBSD Support
Libiocage is not able to fetch and update HardenedBSD releases yet.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/48614369-hardenedbsd-support?utm_campaign=plugin&utm_content=tracker%2F70027116&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F70027116&utm_medium=issues&utm_source=github).
</bountysource-plugin>

gronke:
closes #27 

## Changes
- Fetching HardenedBSD-Releases
- Updating HardenedBSD-Releases
- helpers.exec_iter returns iterable/streamable stdout
- minor other improvements and documentation

### DIFF
--- a/libiocage/cli/fetch.py
+++ b/libiocage/cli/fetch.py
@@ -80,11 +80,14 @@ def cli(ctx, **kwargs):
     logger = ctx.parent.logger
     logger.print_level = kwargs["log_level"]
     host = libiocage.lib.Host.Host(logger=logger)
-    prompts = libiocage.lib.Prompts.Prompts(host=host)
+    prompts = libiocage.lib.Prompts.Prompts(host=host, logger=logger)
 
     release_input = kwargs["release"]
     if release_input is None:
-        release = prompts.release()
+        try:
+            release = prompts.release()
+        except DefaultReleaseNotFound:
+            exit(1)
     else:
         try:
             release = libiocage.lib.Release.Release(
--- a/libiocage/cli/fetch.py
+++ b/libiocage/cli/fetch.py
@@ -28,6 +28,7 @@ import libiocage.lib.Host
 import libiocage.lib.Logger
 import libiocage.lib.Prompts
 import libiocage.lib.Release
+import libiocage.lib.errors
 
 __rootcmd__ = True
 
@@ -86,7 +87,7 @@ def cli(ctx, **kwargs):
     if release_input is None:
         try:
             release = prompts.release()
-        except DefaultReleaseNotFound:
+        except libiocage.lib.errors.DefaultReleaseNotFound:
             exit(1)
     else:
         try:
--- a/libiocage/lib/Distribution.py
+++ b/libiocage/lib/Distribution.py
@@ -21,7 +21,7 @@ class Distribution:
     def __init__(self, host, zfs=None, logger=None):
         libiocage.lib.helpers.init_logger(self, logger)
         libiocage.lib.helpers.init_zfs(self, zfs)
-        self.host = host
+        libiocage.lib.helpers.init_host(self, host)
         self.available_releases = None
         self.zfs = zfs
         self.logger = logger
@@ -56,32 +56,57 @@ class Distribution:
 
     def fetch_releases(self):
 
+        self.logger.spam(f"Fetching release list from '{self.mirror_url}'")
+
         resource = urllib.request.urlopen(self.mirror_url)
         charset = resource.headers.get_content_charset()
         response = resource.read().decode(charset if charset else "UTF-8")
 
-        available_releases = list(map(lambda x: libiocage.lib.Release.Release(
-            name=x,
-            host=self.host,
-            zfs=self.zfs,
-            logger=self.logger
-        ), self._parse_links(response)))
+        found_releases = self._parse_links(response)
 
         available_releases = sorted(
-            available_releases,
-            key=lambda x: float(x.name.partition("-")[0])
+            map( # map long HardenedBSD release names
+                self._map_available_release,
+                filter( # filter out other CPU architectures on HardenedBSD
+                    self._filter_available_releases,
+                    found_releases
+                )
+            )
         )
 
-        self.available_releases = available_releases
-        return available_releases
+        self.available_releases = list(map(
+            lambda x: libiocage.lib.Release.Release(
+                name=x,
+                host=self.host,
+                zfs=self.zfs,
+                logger=self.logger
+            ),
+            available_releases
+        ))
+        return self.available_releases
+
+    def _map_available_release(self, release_name):
+        if self.name == "HardenedBSD":
+            # e.g. HardenedBSD-11-STABLE-libressl-amd64-LATEST
+            return "-".join(release_name.split("-")[1:-2])
+        return release_name        
+
+    def _filter_available_releases(self, release_name):
+        if self.name != "HardenedBSD":
+            return True
+        arch = release_name.split("-")[-2:][0]
+        return self.host.processor == arch
 
     def get_release_trunk_file_url(self, release, filename):
 
         if self.host.distribution.name == "HardenedBSD":
-            # ToDo: implement HardenedBSD release updates
-            raise libiocage.lib.errors.MissingFeature(
-                "Updates of HardenedBSD releases not supported yet"
-            )
+
+            return "/".join([
+                "https://raw.githubusercontent.com/HardenedBSD/hardenedBSD",
+                release.hbds_release_branch,
+                filename
+            ])
+
         elif self.host.distribution.name == "FreeBSD":
 
             if release.name == "11.0-RELEASE":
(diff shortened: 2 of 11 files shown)


libiocage/lib/Release.py
```diff
@@ -36,8 +37,8 @@ def __init__(self, name=None,
         libiocage.lib.helpers.init_zfs(self, zfs)
         libiocage.lib.helpers.init_host(self, host)
 
-        if not libiocage.lib.helpers.validate_name(name):
-            raise NameError(f"Invalid 'name' for Release: '{name}'")
+        # if not libiocage.lib.helpers.validate_name(name):
+        #     raise NameError(f"Invalid 'name' for Release: '{name}'")
```
skarekrow: Debug?

libiocage/lib/Distribution.py
```diff
@@ -56,32 +56,57 @@ def hash_file(self):
...
+        if self.name == "HardenedBSD":
+            # e.g. HardenedBSD-11-STABLE-libressl-amd64-LATEST
+            return "-".join(release_name.split("-")[1:-2])
+        return release_name        
+
+    def _filter_available_releases(self, release_name):
+        if self.name != "HardenedBSD":
+            return True
+        arch = release_name.split("-")[-2:][0]
+        return self.host.processor == arch
```
skarekrow: Is this actually valid syntax? `self.host.arch` isn't defined in `__init__`

libiocage/cli/fetch.py
```diff
@@ -80,11 +80,14 @@ def cli(ctx, **kwargs):
...
     host = libiocage.lib.Host.Host(logger=logger)
-    prompts = libiocage.lib.Prompts.Prompts(host=host)
+    prompts = libiocage.lib.Prompts.Prompts(host=host, logger=logger)
 
     release_input = kwargs["release"]
     if release_input is None:
-        release = prompts.release()
+        try:
+            release = prompts.release()
+        except DefaultReleaseNotFound:
```
skarekrow: This should be `libiocage.lib.errors.DefaultReleaseNotFound`, and you need to import that in `fetch`.

libiocage/lib/Prompts.py
```diff
@@ -17,10 +18,19 @@ def release(self):
...
+        if default is not None:
+            default_release = available_releases[default]
+            selection = input(f"Release ({default_release.name}) [{default}]: ")
+        else:
+            default_release = None
+            selection = input("Your selection: ")
 
         if selection == "":
+            if default_release is None:
+                raise libiocage.lib.errors.DefaultReleaseNotFound(
```
skarekrow: Missing an import for this.

libiocage/lib/Prompts.py
```diff
@@ -17,10 +18,19 @@ def release(self):
...
+                    host_release_name=self.host.release_version,
```
skarekrow: Where is this defined? I don't see it, or `self.host.distribution.releasesc` (above) in `__init__`

libiocage/lib/Release.py
```diff
@@ -310,75 +369,129 @@ def update(self):
...
-            "release"           : self.name,
-            "securelevel"       : "0"
-        },
+                "uuid"              : str(uuid.uuid4()),
+                "basejail"          : False,
+                "allow_mount_nullfs": "1",
+                "release"           : self.name,
+                "securelevel"       : "0"
+            },
+            new=True,
```
skarekrow: Why are we creating a new jail for `update`?

libiocage/lib/Release.py
```diff
@@ -310,75 +369,129 @@ def update(self):
...
             logger=self.logger,
             zfs=self.zfs,
             host=self.host
```
skarekrow: None of these are actually defined from what I can see.

libiocage/lib/Release.py
```diff
@@ -310,75 +369,129 @@ def update(self):
...
+
+            for stdout_line in stdout:
+                self.logger.verbose(stdout_line.strip("\n"), indent=1)
+
+            self.logger.debug(f"Update of release '{self.name}' finished")
+            changed = True
+
+        except:
+
+            raise
```
skarekrow: Redundant raise

libiocage/lib/Release.py
```diff
@@ -310,75 +369,129 @@ def update(self):
...
+                changed = self._update_hbsd_jail(jail)
+            else:
+                changed = self._update_freebsd_jail(jail)
+        except:
+            self.logger.verbose(
+                "There was an error updating the Jail - reverting the changes"
+            )
+            jail.stop(force=True)
+            self.zfs.get_snapshot(snapshot_name).rollback(force=True)
+            raise
```
skarekrow: Perhaps we should have a class here?

libiocage/lib/Release.py
```diff
@@ -310,75 +369,129 @@ def update(self):
...
+                "-j",
+                jail.identifier,
+                "-V"
+            ], logger=self.logger)
+
+            for stdout_line in stdout:
+                self.logger.verbose(stdout_line.strip("\n"), indent=1)
+
+            self.logger.debug(f"Update of release '{self.name}' finished")
+            changed = True
```
skarekrow: Why define `changed`, as you never set it to anything but true, or raise. Just return `True`

libiocage/lib/Release.py
```diff
@@ -310,75 +369,129 @@ def update(self):
...
+                    release_name=self.name,
+                    reason=(
+                        "freebsd-update.sh exited "
+                        f"with returncode {child.returncode}"
+                    ),
+                    logger=self.logger
+                )
+        else:
+            self.logger.debug(f"Update of release '{self.name}' finished")
+            changed = True
```
skarekrow: Same here, why set `changed` at all? It can only ever be true or we raise an exception.

libiocage/lib/Release.py
```diff
@@ -310,75 +369,129 @@ def update(self):
...
+        child, stdout, stderr = jail.exec([
+            "/var/db/freebsd-update/freebsd-update.sh",
+            "-d",
+            "/var/db/freebsd-update",
+            "-f",
+            "/var/db/freebsd-update/freebsd-update.conf",
+            "install"
+        ], ignore_error=True, logger=self.logger)
+
+        if child.returncode == 1:
```
skarekrow: Many scripts like to raise on anything not zero for issues, not always one. This check should be reversed to `if child.returncode != 0`, de-indent and raise in that block instead.

libiocage/lib/Release.py
gronke: We're doing the same for freebsd. It doesn't cost much to start a jail from a release and both freebsd-update.sh and hbsd-update support mechanisms that allow us to do that. On FreeBSD it's even more elegant because it can fetch updates for offline usage.

Anything wrong with that?

libiocage/lib/Release.py
gronke: look out for `libiocage.lib.helpers.init_`. That's where they are initialized.

skarekrow: Then in `__init__` we should have something similar to `self.logger, self.zfs, self.host = libiocage.lib.helpers.init_`, as that's not only non-obvious, it requires chasing context around when debugging things.

libiocage/lib/Release.py
gronke: Yes, had to comment the lines out until a solution to shorten the long HardenedBSD release names. Thanks for reminding me to re-enable those lines.

libiocage/lib/Release.py
gronke: If it improves the readability we could slightly change the interface of our init_ helpers to return the existing or new initialized instance and manually set it, e.g.

```python
self.logger = libiocage.lib.helpers.init_logger(logger)
```

instead of

```python
libiocage.lib.helpers.init_logger(self, logger)
```

We should tackle it in a general refactoring iteration. Do you want me to create a ticket for that?

libiocage/lib/Release.py
gronke: The whole try/except block here is only there to ensure proper teardown in case of a failure. This means when anything goes wrong, we stop the helper jail and roll back the dataset. So the kind of exception that occurred does not matter here.

libiocage/lib/Release.py
gronke: https://github.com/iocage/libiocage/issues/31 most certainly requires action after a release was updated. Left a ToDo note in the code that we need to parse the hbsd-update output and see if changes were applied to the release.

libiocage/cli/fetch.py
gronke: Yeah, thanks! Haven't had issues, because libiocage.lib.errors was imported earlier already. Code that accidentally works.. hm.

gronke:
Merged after addressing @skarekrow's review.

Issue #15, opened 2017-08-22
Feature: DHCP


gronke:
Jails should be allowed to set `ip4_addr=vnet0|dhcp` to obtain an IPv4 address via DHCP on the target interface `vnet0`.

DHCP requires access to /dev/bpf, which needs to be granted via udev rule on the host.

## Requirements
An acceptable solution:
- [ ] creates a set of udev rules per jail
- [ ] removes jail udev rules when stopping a jail
- when `dhcp` is set on any interface in ip4_addr
  - [ ] enables BPF in the jail's udev rule before starting the ip4_addr
- when `dhcp` is not set on any interface in ip4_addr
  - [ ] disables BPF in the jail's udev rule before starting the jail

skarekrow:
This is overkill, and differs from the implementation in iocage currently. The current implementation uses `dhcp` and `bpf` properties as some users wanted to only enable bpf, but not use dhcp. So I think we should migrate any `dhcp` users to the supplied `ip4_addr` syntax here, and add a `bpf` property that is `on` when dhcp is present.

There is no reason to have a udev rule per jail for this however. All jails use the ruleset 4 currently, which is sufficient. We can create just one additional ruleset (exactly how iocage does it now) and have any DHCP jails use that. There isn't a reason to remove things from the devfs rules, that's unnecessary writing to the HDD that we can avoid.

gronke:
> properties as some users wanted to only enable bpf, but not use dhcp

That would be another feature. Indeed it affects the requirements of this one, but it should be tackled in a separate ticket.

> There is no reason to have a udev rule per jail for this however

Maybe a jail want to combine udev rules. It's better to have tailored rules than opening up permissions across all jails.

> that's unnecessary writing to the HDD that we can avoid

Compared to the other operations involved during startup this seems marginal and there are multiple opportunities to optimize this in case it becomes a bottleneck. For example storing all possible combinations that can be activated and giving each a separate id.

skarekrow:
> Maybe a jail want to combine udev rules. It's better to have tailored rules than opening up permissions across all jails.

That doesn't work how you expect. Stacking rules are ignored per the manpage:
```
include ruleset
[...]
Include commands in the referenced ruleset are not resolved.
```

@rexpo sent me a fantastic email detailing this, you can find a forum thread about it here: https://forums.freebsd.org/threads/46091/

[...]

PR #32, opened 2017-08-27
DHCP and dynamic devfs rules

gronke:
implements #15 

- use dhclient to assign a nic's address when `ip4_addr="vnet0|dhcp"` is used
- devfs.rules management tool:
  - read system default rules from `/etc/defaults/devfs.rules`
  - read existing rules from `/etc/devfs.rules` (if the file exists)
  - allows to compare rulesets by their content (lines)
  - allows forking rules from a base rule (used to even extend user specified rulesets `devfs_ruleset="123"` with bpf if necessary)
  - finds the next available devfs rule number
  - leaves comments and whitespace lines untouched
  - recognizes changes and restarts the devfs service when changes were applied
- allows users to specify the devfs rule by name (`devfs_rule="my_devfs_rule"`) instead of selection by number only

### DIFF
--- a/libiocage/lib/DevfsRules.py
+++ b/libiocage/lib/DevfsRules.py
@@ -1,0 +1,382 @@
+import re
+import os.path
+
+import libiocage.lib.errors
+import libiocage.lib.helpers
+
+
+class DevfsRulesFilter:
+
+    def __init__(self, source, active_filters=[]):
+        self.source = source
+        self.active_filters = active_filters
+
+    def with_rule(self, rule):
+        return DevfsRulesFilter(filter(
+            lambda x: x.has_rule(rule),
+            self.source
+        ), self.active_filters + [rule])
+
+    def with_include(self, rule_name):
+        rule = f"add include ${rule_name}"
+        return self.with_rule(rule)
+
+    def create_ruleset(self, ruleset_name, ruleset_number):
+        """
+        Create a ruleset from the provided filters
+
+        Args:
+
+            ruleset_name (string):
+                Name of the rule that get's created
+
+            ruleset_number (int):
+                Number of the rule that get's created
+        """
+
+        ruleset = DevfsRuleset(ruleset_name, ruleset_number)
+        for line in self.active_filters:
+            ruleset.append(line)
+        return ruleset
+
+
+class DevfsRuleset(list):
+    """
+    Representation of a devfs ruleset in the devfs.rules file
+
+    DevfsRuleset instances behave like standard lists and can store strings
+    that multiple lines (string)
+    """
+
+    PATTERN = re.compile(r"""^\[(?P<name>[a-z](?:[a-z0-9\-_]*[a-z0-9])?)=
+                (?P<number>[0-9]+)\]\s*(?:\#\s*(?P<comment>.*))?$""", re.X)
+
+    def __init__(self, value=None, number=None, comment=None):
+        """
+        Initialize DevfsRuleset
+
+        Args:
+
+            value (string): (optional)
+                If specified in combination with a number, this parameter is
+                interpreted as the ruleset name. Otherwise it is parsed with
+                the expectation to find a [<name>=<number>] ruleset definition.
+
+                When value is not specified or None, DevfsRuleset is assumed
+                to be new or unspecified (cannot be exported before name and
+                number were assigned at a later time)
+
+            number (int): (optional)
+                The number of the ruleset. Must be specified to export the
+                ruleset, but is like the name not required to compare the
+                ruleset with another
+        """
+
+        # when only one argument is passed, it's a line that need to be parsed
+        if value is None and number is None:
+            # name and number will be assigned later
+            name = None
+        elif number is None:
+            name, number, comment = self._parse_line(value)
+        else:
+            name = int(value)
+
+        self.name = name
+        self.number = number
+        self.comment = comment
+        list.__init__(self)
+
+    def has_rule(self, rule):
+        """
+        Returns true if the rule is part of the current ruleset
+
+        Args:
+
+            rule (string):
+                The rule string to be compared with current rules of the
+                ruleset instance
+        """
+        return rule in self
+
+    def append(self, rule):
+        if rule not in self:
+            list.append(self, rule)
+
+    def clone(self, source_ruleset):
+        """
+        Clones the rules from another ruleset
+
+        Args:
+
+            source_ruleset (libiocage.lib.DevfsRules.DevfsRuleset):
+                Ruleset to copy all rules from
+        """
+        for rule in source_ruleset:
+            self.append(rule)
+
+    def _parse_line(self, line):
+
+        # marks beginning of a new ruleset
+        ruleset_match = re.search(DevfsRuleset.PATTERN, line)
+        if ruleset_match is not None:
+            name = str(ruleset_match.group("name"))
+            number = int(ruleset_match.group("number"))
+            comment = ruleset_match.group("comment")
+            return name, number, comment
+
+        raise SyntaxError("DevfsRuleset line parsing failed")
+
+    def __str__(self):
+        ruleset_line = f"[{self.name}={self.number}]"
+        if self.comment is not None:
+            ruleset_line += f" # {self.comment}"
+        output = [ruleset_line] + [str(x) for x in self]
+        return "\n".join(output) + "\n"
+
+
+class DevfsRules(list):
+    """
+    Abstraction for the hosts /etc/devfs.rules
+
+    Read and edit devfs rules in a programmatic way.
+    Restarts devfs service after applying changes.
+    """
+
+    def __init__(self, rules_file="/etc/devfs.rules", logger=None):
+        """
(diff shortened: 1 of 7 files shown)


libiocage/lib/DevfsRules.py
```diff
@@ -0,0 +1,367 @@
...
+
+class DevfsRuleset(list):
+    """
+    Representation of a devfs ruleset in the devfs.rules file
+
+    DevfsRuleset instances behave like standard lists and can store strings
+    that  multiple lines (string)
+    """
+
+    PATTERN = r"^\[(?P<name>[a-z](?:[a-z0-9\-_]*[a-z0-9])?)=(?P<number>[0-9]+)\]\s*(?:#\s*(?P<comment>.*))?$"
```
gronke: @araujobsd @skarekrow do you have any idea how to reasonably shorten this line?

@igalic FYI

libiocage/lib/Logger.py
```diff
@@ -138,7 +138,7 @@ def _print(self, message, level, jail=None, indent=0):
 
     def _indent(self, message, level):
         indent = Logger.INDENT_PREFIX * level
-        return f"{indent}{message}"
+        return "\n".join(map(lambda x: f"{indent}{x}", message.split("\n")))
```
gronke: Indention support for multi-line logs

libiocage/lib/DevfsRules.py
igalic: the `/x`, or verbose flag, might be better than shortening it, https://docs.python.org/3/library/re.html#re.VERBOSE

gronke: looks good. will apply that!

libiocage/lib/DevfsRules.py
```diff
@@ -0,0 +1,381 @@
...
+class DevfsRuleset(list):
+    """
+    Representation of a devfs ruleset in the devfs.rules file
+
+    DevfsRuleset instances behave like standard lists and can store strings
+    that  multiple lines (string)
+    """
+
+    PATTERN = r"""^\[(?P<name>[a-z](?:[a-z0-9\-_]*[a-z0-9])?)=
+                (?P<number>[0-9]+)\]\s*(?:#\s*(?P<comment>.*))?$"""
```
igalic: you don't have to escape the `#`??

libiocage/lib/DevfsRules.py
```diff
@@ -0,0 +1,381 @@
...
+            source_ruleset (libiocage.lib.DevfsRules.DevfsRuleset):
+                Ruleset to copy all rules from
+        """
+        for rule in source_ruleset:
+            self.append(rule)
+
+    def _parse_line(self, line):
+
+        # marks beginning of a new ruleset
+        ruleset_match = re.search(DevfsRuleset.PATTERN, line)
```
igalic: how come you don't have to compile the pattern to tell it that it's `/x`?

libiocage/lib/DevfsRules.py
```diff
@@ -0,0 +1,381 @@
...
+        for line in self.active_filters:
+            ruleset.append(line)
+        return ruleset
+
+class DevfsRuleset(list):
+    """
+    Representation of a devfs ruleset in the devfs.rules file
+
+    DevfsRuleset instances behave like standard lists and can store strings
+    that  multiple lines (string)
```
skarekrow: Extra space here

libiocage/lib/DevfsRules.py
skarekrow: He shouldn't have to, it's in a docstring `"""`

libiocage/lib/DevfsRules.py
```diff
@@ -0,0 +1,381 @@
...
+
+        next_line_index = len(self)
+
+        if ruleset is None or isinstance(ruleset, str):
+            list.append(self, ruleset)
+            if is_system_rule is True:
+                self._system_rule_lines.append(next_line_index)
+            return ruleset
+
+        if ruleset.name in self._ruleset_name_index.keys():
```
skarekrow: I think this and the last if below can be combined into an or

libiocage/lib/DevfsRules.py
```diff
@@ -0,0 +1,381 @@
...
+
+    def find_by_number(self, rule_number):
+        return self._find_by_index(rule_number, self._ruleset_number_index)
+
+    def _find_by_index(self, rule_name, index):
+        return self[index[rule_name]]
+
+    @property
+    def default_rules_file(self):
+        return "/etc/defaults/devfs.rules"
```
skarekrow: Why are we using `/etc/defaults/devfs.rules` instead of `/etc/devfs.rules`?

libiocage/lib/Host.py
```diff
@@ -21,9 +21,22 @@ def __init__(self, root_dataset=None, zfs=None, logger=None):
...
+        self._devfs = None
         self.releases_dataset = None
 
     @property
+    def devfs(self):
+        """
+        Lazy-loaded DevfsRules instance
+        """
+        if self._devfs is None:
+            import libiocage.lib.DevfsRules
```
skarekrow: We should import at the top, this is against PEP8 (https://www.python.org/dev/peps/pep-0008/#imports)

libiocage/lib/Jail.py
```diff
@@ -369,6 +370,56 @@ def _destroy_jail(self):
...
 
+    @property
+    def _dhcp_enabled(self):
+        """
+        True if any ip4_addr uses DHCP
+        """
+        if self.config["ip4_addr"] is None:
+            return False
+
+        return ("dhcp" in self.config["ip4_addr"].networks)
```
skarekrow: Redundant parenthesis

libiocage/lib/DevfsRules.py
gronke: We are using both - "system" rules are loaded from `/etc/defaults/devfs.rules`.

That's important to be able to inherit rules from them. Have you noticed the real devfs_rule number provided by the user may be altered? In case the combination of iocage-required devfs rules and the user supplied rule (default `4`) has not been seen in this exact combination, a new rule is added permanently. When this happens, the devfs service is restarted.

libiocage/lib/Host.py
gronke: Do you know a way to still lazy load the module? It's not required most of the scenarios using libiocage, so I thought delaying the loading can be benifitial for all other tasks. The performance penalty will probably not be noticeable, so in case you don't know a way, I'm up to move it to the files header.

libiocage/lib/Jail.py
gronke: Hehehe.. But it's a bool :neckbeard: I'm so used to have operators that resolve to a boolean value in parenthesis.

libiocage/lib/DevfsRules.py
gronke: I do 🙈 ... fixed

libiocage/lib/Host.py
skarekrow: I can't think of one offhand, but yeah the penalty will be so minimal. It's nice to have all our imports in one place.

libiocage/lib/DevfsRules.py
gronke: Could be, but we need to still differentiate the message, so I'd think it would not look much cleaner after moving around. If you see more potential, please poke me again ;)

libiocage/lib/DevfsRules.py
```diff
@@ -0,0 +1,382 @@
...
+
+        # marks beginning of a new ruleset
+        ruleset_match = re.search(DevfsRuleset.PATTERN, line)
+        if ruleset_match is not None:
+            name = str(ruleset_match.group("name"))
+            number = int(ruleset_match.group("number"))
+            comment = ruleset_match.group("comment")
+            return name, number, comment
+
+        raise SyntaxError("DevfsRuleset line parsing failed")
```
igalic: maybe turn this if around, raise early, unident the happy path

libiocage/lib/DevfsRules.py
```diff
@@ -0,0 +1,382 @@
...
+        self._system_rule_lines = []
+
+        list.__init__(self)
+
+        # will automatically read from file - needs to be the last item
+        self.rules_file = rules_file
+
+    def append(self, ruleset, is_system_rule=False):
+        """
+        Add a DevfsRuleset to the list
```
igalic: this list contains different types.
doesn't that make it a tuple?

libiocage/lib/DevfsRules.py
```diff
@@ -0,0 +1,382 @@
...
+
+            # the first item must be a ruleset
+            if current_ruleset is None:
+                raise libiocage.lib.errors.InvalidDevfsRulesSyntax(
+                    devfs_rules_file=self.rules_file,
+                    reason="Rules must follow a ruleset declaration",
+                    logger=self.logger
+                )
+
+            current_ruleset.append(line)
```
igalic: i think I'm just in general very confused how this parsing works.

Issue #26, opened 2017-08-24
Feature: Default config


gronke:
Currently libiocage has hardcoded default config properties. In iocage it should be possible to define global defaults by executing `iocage set foo=bar default`.

By providing a full list of defaults, each jails configuration only needs to contain the delta. When initializing the defaults, the global defaults should be read and override the hardcoded ones.

Even though it does not have influence on the way user defined globals work, it seems cleaner to export the hardcoded values to a json file instead as well.

skarekrow:
This currently exists in iocage, I would suggest that we use the defaults as a base, and have the jails use a differential configuration on top of that. Instead of using it as a seed, like the current implementation in iocage does.

PR #35, opened 2017-08-27
Default Config

gronke:
implements #26

- [x] load default configuration from default.json
- [x] assumes default.json is stored in zpool `<ACTIVE_POOL>/iocage`
- [ ] can modify default.json

### Additional Changes

- cleaned up list of system defaults
- more comments

### DIFF
--- a/libiocage/lib/JailConfig.py
+++ b/libiocage/lib/JailConfig.py
@@ -7,6 +7,7 @@ import libiocage.lib.JailConfigJSON
 import libiocage.lib.JailConfigLegacy
 import libiocage.lib.JailConfigResolver
 import libiocage.lib.JailConfigZFS
+import libiocage.lib.JailConfigDefaults
 import libiocage.lib.errors
 import libiocage.lib.helpers
 
@@ -45,7 +46,12 @@ class JailConfig(dict, object):
 
     """
 
-    def __init__(self, data={}, jail=None, logger=None, new=False):
+    def __init__(self,
+                 data={},
+                 jail=None,
+                 logger=None,
+                 new=False,
+                 defaults_file=None):
 
         dict.__init__(self)
 
@@ -59,7 +65,9 @@ class JailConfig(dict, object):
         if jail:
             self.jail = jail
             fstab = libiocage.lib.JailConfigFstab.JailConfigFstab(
-                jail=jail, logger=self.logger)
+                jail=jail,
+                logger=self.logger
+            )
             self.fstab = fstab
         else:
             self.jail = None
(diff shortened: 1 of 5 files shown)


libiocage/cli/start.py
```diff
@@ -51,6 +51,7 @@ def cli(ctx, rc, jails, log_level):
         try:
             jail.start()
         except Exception:
+            raise
```
skarekrow: `exit(1)` will never be reached.

libiocage/lib/JailConfig.py
```diff
@@ -340,16 +373,13 @@ def _set_jail_zfs(self, value, **kwargs):
             value, true="on", false="off")
 
     def _default_jail_zfs(self):
-        # if self.data["jail_zfs"] does not explicitly exist, _get_jail_zfs
-        # would raise
+        # if self.data["jail_zfs"] does not explicitly exist,
+        # _get_jail_zfs would raise
```
skarekrow: Let's make this a docstring, two comments at the top look ugly

libiocage/lib/JailConfig.py
```diff
@@ -370,18 +400,28 @@ def _get_cloned_release(self):
...
-        return self.data["basejail_type"]
 
-    def _default_basejail_type(self):
+        # first see if basejail_type was explicitly set
+        try:
+            return self.data["basejail_type"]
+        except:
+            pass
+
+        # if it was not, the default for is 'nullfs' if the jail is a basejail
```
skarekrow: `If it was not, the default for a basejail type is nullfs` reads better

libiocage/lib/JailConfigJSON.py
```diff
@@ -24,7 +24,9 @@ def read(self):
 
     def read_data(self):
         with open(JailConfigJSON.__get_config_json_path(self), "r") as conf:
-            return json.load(conf)
+            data = json.load(conf)
+            conf.close()
```
skarekrow: The `with` statement already closes once you leave it, that shouldn't be necessary.

libiocage/lib/errors.py
```diff
@@ -100,14 +100,19 @@ def __init__(self, value, **kwargs):
...
 
 
-class JailConfigNotFound(Exception):
-    # This is a silent error internally used
-
+class JailConfigNotFound(IocageException):
     def __init__(self, config_type, *args, **kwargs):
         msg = f"Could not read {config_type} config"
+        # This is a silent error internally used
         Exception.__init__(self, msg, *args, **kwargs)
```
skarekrow: This isn't correct, `__init__` only accepts `*args` and `**kwargs`. If we ever used this internally, it wouldn't function.

gronke: There is a problem here - right. But we're seeing different ones here. I would have removed `*args, **kwargs` from Exception.__init__, because we only want to raise this message and nothing else.

PR #36, opened 2017-08-27
Fix all flake8 issues, remove unused import and duplicate error classes.

skarekrow:
Changes
======
- Remove all the flake8 violations
- Remove unused import
- Remove duplicate error class

### DIFF
--- a/libiocage/lib/JailConfig.py
+++ b/libiocage/lib/JailConfig.py
@@ -41,7 +41,7 @@ class JailConfig(dict, object):
 
     Special Properties:
 
-        Special properties are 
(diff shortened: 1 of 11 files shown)


libiocage/lib/JailConfig.py
```diff
@@ -41,7 +41,7 @@ class JailConfig(dict, object):
 
     Special Properties:
 
-        Special properties are 
+        Special properties are
```
gronke: The author suddenly stopped writing *couch*

skarekrow: Haha

Issue #16, opened 2017-08-22
Have the same user output as iocage


skarekrow:
start: Little output, should look the same. JID can be included in the same style.

stop: Little output, should look the same.

There are likely more, but just creating this as a placeholder so when this becomes iocage, the user will not notice a difference. This is important for scripting and other things we aren't necessarily privy of.

gronke:
@skarekrow can you please add samples of expected output?

skarekrow:
Sure.

`iocage start foo`:
```
* Starting foo
  + Started OK
  + Starting services OK
```

`iocage start bar`:
```
* Starting bar
  + Started OK
  + Configuring VNET OK
  + Starting services OK
```

`iocage stop foo`:
```
* Stopping foo
  + Running prestop OK
  + Stopping services OK
  + Removing jail process OK
  + Running poststop OK
```

`iocage stop bar`:
```
* Stopping bar
  + Running prestop OK
  + Stopping services OK
  + Tearing down VNET OK
  + Removing jail process OK
  + Running poststop OK
```

Basically, run `iocage` and mirror that output, most messages have been chosen for a good reason, like the fdescfs warning we should bring back (massive slow down if your host has lots of ram and you aren't using it.)

gronke:
We need IPC between the cli and the lib to achieve that. I was thinking about using a socket, so that we can easily expose that for to interfaces as well via WebSocket proxy.

igalic:
> we need IPC

let me stop you right there.  😛

skarekrow:
@gronke Yeah that's why I put this as a ticket, it will require more effort instead of the monolithic approach iocage used. 😆

gronke:
Generators seem to be the technology we want to use here. Specific public methods should optionally allow to return a generator that pushes update events from the lib to the consumer (in our case the CLI). Will prepare a proof of concept to discuss.

gronke:
With merging the generators branch the requirement to make the iocage cli look the same is now up to the CLI implementation, not the library -> Closing

PR #37, opened 2017-08-27
Commandline <> Lib communication using generators

gronke:
implements #16 

Specific class methods will allow to set a `yields=True` argument that will turn it into a Python generator. This generator will push status updates in form of `libiocage.lib.events.*`

### Demo Snippets

This are some demos that demonstrate generators. They don't aim to be the exact output we will provide to the users.

Fetch: https://asciinema.org/a/CSpa7Nw9KMXrD4MAUo9TItNlI
List: https://asciinema.org/a/Azzsm7RBOdRCunwzQpcFcio3v
Bulk-Set: https://asciinema.org/a/8bUcbXAk3MPwfcOkq7RNoNNyG

### DIFF
--- a/libiocage/cli/__init__.py
+++ b/libiocage/cli/__init__.py
@@ -60,6 +60,57 @@ except su.CalledProcessError:
     exit(1)
 
 
+def print_events(generator):
+    lines = {}
+    last_event = None
+    for event in generator:
+
+        if event.identifier is None:
+            identifier = "generic"
+        else:
+            identifier = event.identifier
+        
+        if event.type not in lines:
+            lines[event.type] = {}
+
+        # output fragments
+        running_indicator = "+" if (event.done or event.skipped) else "-"
+        name = event.type
+        if event.identifier is not None:
+            name += f"@{event.identifier}"
+
+        output = f"[{running_indicator}] {name}: "
+
+        if event.message is not None:
+            output += event.message
+        else:
+            output += event.get_state_string(
+                done="OK",
+                error="failed",
+                skipped="skipped",
+                pending="..."
+            )
+
+        if event.duration is not None:
+            output += " [" + str(round(event.duration, 3)) + "s]"
+
+        indent = event.parent_count
+        # new line or update of previous
+        if identifier not in lines[event.type]:
+            # Indent if previous task is not finished
+            lines[event.type][identifier] = logger.screen(
+                output,
+                indent=event.parent_count
+            )
+        else:
+            lines[event.type][identifier].edit(
+                output,
+                indent=event.parent_count
+            )
+
+        last_event = event
+
+
 class IOCageCLI(click.MultiCommand):
     """
     Iterates in the 'cli' directory and will load any module's cli definition.
@@ -77,7 +128,7 @@ class IOCageCLI(click.MultiCommand):
         return rv
 
     def get_command(self, ctx, name):
-        ctx.logger = logger
+        ctx.print_events = print_events
         try:
             mod = __import__(f"libiocage.cli.{name}", None, None, ["cli"])
 
@@ -96,11 +147,15 @@ class IOCageCLI(click.MultiCommand):
                 pass
             return mod.cli
         except (ImportError, AttributeError):
+            raise
             return
 
-
+@click.option("--log-level", "-d", default=None)
 @click.command(cls=IOCageCLI)
-@click.version_option(version="0.9.10 07/30/2017", prog_name="iocage",
+@click.version_option(version="0.2.11 08/29/2017", prog_name="ioc",
                       message="%(version)s")
-def cli():
+@click.pass_context
+def cli(ctx, log_level):
     """A jail manager."""
+    logger.print_level = log_level
+    ctx.logger = logger
--- a/libiocage/cli/__init__.py
+++ b/libiocage/cli/__init__.py
@@ -62,14 +62,13 @@ except su.CalledProcessError:
 
 def print_events(generator):
     lines = {}
-    last_event = None
     for event in generator:
 
         if event.identifier is None:
             identifier = "generic"
         else:
             identifier = event.identifier
-        
+
         if event.type not in lines:
             lines[event.type] = {}
 
@@ -94,7 +93,6 @@ def print_events(generator):
         if event.duration is not None:
             output += " [" + str(round(event.duration, 3)) + "s]"
 
-        indent = event.parent_count
         # new line or update of previous
         if identifier not in lines[event.type]:
             # Indent if previous task is not finished
@@ -107,8 +105,6 @@ def print_events(generator):
                 output,
                 indent=event.parent_count
             )
-
-        last_event = event
 
 
 class IOCageCLI(click.MultiCommand):
@@ -149,6 +145,7 @@ class IOCageCLI(click.MultiCommand):
         except (ImportError, AttributeError):
             raise
             return
+
 
 @click.option("--log-level", "-d", default=None)
 @click.command(cls=IOCageCLI)
--- a/libiocage/cli/__init__.py
+++ b/libiocage/cli/__init__.py
@@ -85,8 +85,8 @@ def print_events(generator):
         else:
             output += event.get_state_string(
                 done="OK",
-                error="failed",
-                skipped="skipped",
+                error="FAILED",
+                skipped="SKIPPED",
                 pending="..."
             )
 
@@ -143,7 +143,6 @@ class IOCageCLI(click.MultiCommand):
                 pass
             return mod.cli
         except (ImportError, AttributeError):
-            raise
             return
 
 
--- a/libiocage/cli/fetch.py
+++ b/libiocage/cli/fetch.py
@@ -33,19 +33,6 @@ import libiocage.lib.errors
 __rootcmd__ = True
 
 
-# ToDo: remove disabled feature
-# def _prettify_release_names(x):
-#     if x.name == host.release_version:
-#         return f"\033[1m{x.name}\033[0m"
-#     else:
-#         return x.name
-# def release_choice():
-#     version =
-#     return click.Choice(list(map(
-#         _prettify_release_names,
-#         host.distribution.releases
-#     )))
-
 @click.command(context_settings=dict(
     max_content_width=400, ),
     name="fetch", help="Fetch a version of FreeBSD for jail usage or a"
@@ -69,15 +56,10 @@ __rootcmd__ = True
 @click.option("--files", multiple=True,
               help="Specify the files to fetch from the mirror. "
                    "(Deprecared: renamed to --file)")
-@click.option("--log-level", "-d", default=None)
 def cli(ctx, **kwargs):
     logger = ctx.parent.logger
-    logger.print_level = kwargs["log_level"]
     host = libiocage.lib.Host.Host(logger=logger)
     prompts = libiocage.lib.Prompts.Prompts(host=host, logger=logger)
-
-    if kwargs["log_level"] is not None:
-        logger.print_level = kwargs["log_level"]
 
     release_input = kwargs["release"]
     if release_input is None:
@@ -87,7 +69,7 @@ def cli(ctx, **kwargs):
             exit(1)
     else:
         try:
-            release = libiocage.lib.Release.Release(
+            release = libiocage.lib.Release.ReleaseGenerator(
                 name=release_input,
                 host=host,
                 logger=logger
@@ -111,10 +93,10 @@ def cli(ctx, **kwargs):
         exit(1)
 
     fetch_updates = bool(kwargs["fetch_updates"])
-    release.fetch(
+    ctx.parent.print_events(release.fetch(
         update=kwargs["update"],
         fetch_updates=fetch_updates
-    )
+    ))
 
     exit(0)
 
--- a/libiocage/cli/fetch.py
+++ b/libiocage/cli/fetch.py
@@ -25,10 +25,10 @@
 import click
 
 import libiocage.lib.Host
-import libiocage.lib.Logger
 import libiocage.lib.Prompts
 import libiocage.lib.Release
 import libiocage.lib.errors
+
 
 __rootcmd__ = True
 
@@ -59,8 +59,7 @@ __rootcmd__ = True
               # type=release_choice(),
               help="The FreeBSD release to fetch.")
 @click.option("--update/--no-update", "-U/-NU", default=True,
-              help="Decide whether or not to update the fetch to the latest "
-                   "patch level.")
+              help="Update the release to the latest patch level.")
 @click.option("--fetch-updates/--no-fetch-updates", default=True,
               help="Skip fetching release updates")
 # Compat
@@ -71,17 +70,14 @@ __rootcmd__ = True
               help="Specify the files to fetch from the mirror. "
                    "(Deprecared: renamed to --file)")
 @click.option("--log-level", "-d", default=None)
-# @click.option("--auth", "-a", default=None, help="Authentication method for "
-#                                                 "HTTP fetching. Valid "
-#                                                 "values: basic, digest")
-# @click.option("--verify/--noverify", "-V/-NV", default=True,
-#               help="Enable or disable verifying SSL cert for HTTP fetching.")
-# def cli(url, files, release, update):
 def cli(ctx, **kwargs):
     logger = ctx.parent.logger
     logger.print_level = kwargs["log_level"]
     host = libiocage.lib.Host.Host(logger=logger)
     prompts = libiocage.lib.Prompts.Prompts(host=host, logger=logger)
+
+    if kwargs["log_level"] is not None:
+        logger.print_level = kwargs["log_level"]
 
     release_input = kwargs["release"]
     if release_input is None:
@@ -100,9 +96,6 @@ def cli(ctx, **kwargs):
             logger.error(f"Invalid Release '{release_input}'")
             exit(1)
 
-    if kwargs["log_level"] is not None:
-        logger.print_level = kwargs["log_level"]
-
     url_or_files_selected = False
 
     if is_option_enabled(kwargs, "url"):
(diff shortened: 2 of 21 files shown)


libiocage/cli/start.py
```diff
@@ -49,8 +49,11 @@ def cli(ctx, rc, jails, log_level):
     for jail in ioc_jails.list(filters=jails):
         logger.log(f"Starting {jail.humanreadable_name}")
         try:
-            jail.start()
+            for i in jail.start(yields=True):
+                print(f"[+] {i.action} OK")
+
         except Exception:
+            raise
             exit(1)
```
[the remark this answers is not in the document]
gronke: I'm sorry, but this is necessary to see python stacktraces, so I have this enabled until I prepare the branch for the merge. The errors that get swallowed from exit(1) are structural ones, not feature failures, so they are useful for development only. Already told @skarekrow that I'll find a better solution for this, because it lead to quite some confusion. For example this one https://github.com/iocage/libiocage/pull/35#discussion_r135416094

:trollface:

gronke:
Current Status: https://asciinema.org/a/CSpa7Nw9KMXrD4MAUo9TItNlI

libiocage/lib/Release.py
```diff
@@ -332,7 +376,7 @@ def fetch_updates(self):
                             "Components"
                         ))
                         f.truncate()
-                        f.close()
```
skarekrow: Why are you removing this? I don't see it being used anywhere, this will result in leaking fd's

libiocage/lib/Release.py
```diff
@@ -249,27 +247,68 @@ def fetch(self, update=None, fetch_updates=None):
...
 
-        auto_update_on = self.auto_update and update is not False
-        if auto_update_on or update:
-            release_changed = self.update()
+        if update is True:
+            for event in ReleaseGenerator.update(self):
+                if isinstance(event, libiocage.lib.events.IocageEvent):
+                    yield event
+                else:
+                    # the only non-IocageEvent is out return value
```
skarekrow: s/out/our/

libiocage/lib/Logger.py
```diff
@@ -80,70 +113,95 @@ def log(self, *args, **kwargs):
...
 
-    def spam(self,
-             message,
-             jail=None,
-             indent=0):
+        line_number = self.PRINT_HISTORY.index(log_entry)
+        delta = len(self.PRINT_HISTORY) - line_number
 
-        self.log(message, level="spam", jail=jail, indent=indent)
+        # ToDo: Handle redrawig of multiline entries with different line count
```
skarekrow: s/redrawig/redrawing/ Sorry, OCD!

libiocage/lib/Logger.py
```diff
@@ -80,70 +113,95 @@ def log(self, *args, **kwargs):
...
+
+        output = "".join([
+            #"\033[s",            # SCP - Save Cursor Position
+            f"\033[{delta}F",  # CPL - Cursor Previous Line
+            "\r",               # CR - Carriage Return
+            self._indent(log_entry.message, log_entry.indent),
+            "\033[K",           # EL - Erease in Line
```
skarekrow: s/Erease/Erase/

libiocage/lib/Logger.py
```diff
@@ -80,70 +113,95 @@ def log(self, *args, **kwargs):
...
-             jail=None,
-             indent=0):
+        line_number = self.PRINT_HISTORY.index(log_entry)
+        delta = len(self.PRINT_HISTORY) - line_number
 
-        self.log(message, level="spam", jail=jail, indent=indent)
+        # ToDo: Handle redrawig of multiline entries with different line count
+
+        output = "".join([
+            #"\033[s",            # SCP - Save Cursor Position
```
skarekrow: If this is used (assuming you left a comment on it because it was useful), why is it commented out then?

libiocage/lib/Logger.py
```diff
@@ -80,70 +113,95 @@ def log(self, *args, **kwargs):
...
+            f"\033[{delta}F",  # CPL - Cursor Previous Line
+            "\r",               # CR - Carriage Return
+            self._indent(log_entry.message, log_entry.indent),
+            "\033[K",           # EL - Erease in Line
+            "\n" * delta
+            #"\033[s"            # RCP - Restore Cursor Position
```
skarekrow: Same here

libiocage/lib/Logger.py
```diff
@@ -80,70 +113,95 @@ def log(self, *args, **kwargs):
...
+        ])
+
+        sys.stdout.write(output)
+        #print(log_entry.message)
```
skarekrow: Debug print

libiocage/lib/Logger.py
```diff
@@ -80,70 +113,95 @@ def log(self, *args, **kwargs):
...
+    def spam(self, message, indent=0, **kwargs):
+        return self.log(message, level="spam", indent=indent, **kwargs)
 
-    def warn(self,
-             message,
-             jail=None,
-             indent=0):
+    def screen(self, message, indent=0, **kwargs):
+        """
+        Screen does never get printed to log files
```
skarekrow: You can drop `does` for that sentence, then add a s to `get`

libiocage/lib/Jail.py
```diff
@@ -511,7 +552,7 @@ def _launch_jail(self):
 
     def _start_vimage_network(self):
 
-        self.logger.log("Starting VNET/VIMAGE", jail=self)
+        self.logger.debug("Starting VNET/VIMAGE", jail=self)
```
skarekrow: Why debug and not info?

libiocage/cli/fetch.py
```diff
@@ -71,18 +70,15 @@
               help="Specify the files to fetch from the mirror. "
                    "(Deprecared: renamed to --file)")
 @click.option("--log-level", "-d", default=None)
-# @click.option("--auth", "-a", default=None, help="Authentication method for "
-#                                                 "HTTP fetching. Valid "
-#                                                 "values: basic, digest")
-# @click.option("--verify/--noverify", "-V/-NV", default=True,
-#               help="Enable or disable verifying SSL cert for HTTP fetching.")
-# def cli(url, files, release, update):
```
skarekrow: Why are we removing current CLI flags?

libiocage/lib/Logger.py
gronke: I was playing with various techniques to alter already printed output. Didn't get it to work as expected, but it was the most promising solution when it comes to multi-line editing

libiocage/lib/Jail.py
gronke: The library is not supposed to print. Info/Log gets printed by default and we want all stdout to be generated from the event responses.

libiocage/cli/fetch.py
gronke: Because they were not implemented yet, so I removed it to clean up the file. Most probably https://github.com/iocage/libiocage/issues/43 will cause need to re-apply and use this option. I've updated the CLI criteria for signature checking, so that it's not forgotten.

libiocage/lib/Release.py
gronke: `with` is already taking care of that. See http://effbot.org/zone/python-with-statement.htm

skarekrow: I know how `with` works ;) I commented on that in a different ticket. I just forgot to click that little expand thing.

libiocage/cli/list.py
```diff
@@ -68,52 +71,120 @@ def cli(ctx, dataset_type, header, _long, remote, plugins,
...
-    return
 
+def _lookup_jail_value(
+    jail: libiocage.lib.Jail.JailGenerator,
+    key: str
+) -> str:
 
-def _lookup_jail_value(jail, key):
-    if key in libiocage.lib.Jails.Jails.JAIL_KEYS:
+    # ToDo: Move this into lib/Jails ?
```
skarekrow: This I think better belongs in `helpers`

libiocage/cli/start.py
```diff
@@ -35,23 +35,32 @@
...
+    for jail in ioc_jails:
         logger.log(f"Starting {jail.humanreadable_name}")
         try:
-            jail.start()
+            ctx.parent.print_events(jail.start())
+
         except Exception:
+            failed_jails.append(jail)
+            continue
             exit(1)
```
skarekrow: `exit` after `continue`?

libiocage/cli/start.py
```diff
@@ -35,23 +35,32 @@
...
 
         logger.log(f"{jail.humanreadable_name} running as JID {jail.jid}")
-        exit(0)
+
+    if len(failed_jails) > 0:
```
skarekrow: I think these two can be a one liner

libiocage/cli/stop.py
```diff
@@ -45,15 +45,22 @@ def cli(ctx, rc, log_level, force, jails):
...
         except:
-            exit(1)
+            failed_jails.append(jail)
+            continue
 
-        logger.log("done")
-        exit(0)
+        logger.log(f"{jail.name} stopped")
+
+    if len(failed_jails) > 0:
```
skarekrow: Same here

libiocage/lib/Jail.py
```diff
@@ -852,9 +890,26 @@ def __dir__(self):
...
         return list(properties)
+
+
+class Jail(JailGenerator):
+
+    _class_host = libiocage.lib.Host.HostGenerator
+
+    def start(self, *args, **kwargs):
+        return list(JailGenerator.start(self, *args, **kwargs))
+        # libiocage.lib.helpers.print_event_generator(
```
skarekrow: Since this code doesn't exist in master anyways, we should just remove the commented code. Unless you didn't accidentally push it?

libiocage/lib/Jail.py
```diff
@@ -852,9 +890,26 @@ def __dir__(self):
...
+        #     JailGenerator.start(self, *args, **kwargs),
+        #     logger=self.logger
+        # )
+
+    def stop(self, *args, **kwargs):
+        return list(JailGenerator.start(self, *args, **kwargs))
+        # libiocage.lib.helpers.print_event_generator(
```
skarekrow: Meep. Same here

libiocage/lib/JailFilter.py
```diff
@@ -0,0 +1,132 @@
+from typing import List, Union, Iterable
```
skarekrow: Missing iocage license

libiocage/lib/Host.py
```diff
@@ -7,7 +7,10 @@
 import libiocage.lib.helpers
```
skarekrow: Missing iocage license at top

libiocage/lib/Distribution.py
```diff
@@ -8,7 +8,10 @@
 import libiocage.lib.helpers
```
skarekrow: Missing iocage license at top

libiocage/lib/events.py
```diff
@@ -0,0 +1,275 @@
+from timeit import default_timer as timer
```
skarekrow: Missing iocage license

libiocage/lib/Logger.py
```diff
@@ -1,9 +1,38 @@
 import os
```
skarekrow: Missing iocage license

libiocage/lib/Jails.py
```diff
@@ -1,12 +1,13 @@
-import re
+from typing import Generator, Union, Iterable
```
skarekrow: Missing iocage license

libiocage/lib/Jails.py
```diff
@@ -1,12 +1,13 @@
...
 
 import libzfs
 
 import libiocage.lib.Jail
+import libiocage.lib.JailFilter
 import libiocage.lib.helpers
```
skarekrow: Are you running isort like CONTRIBUTING says? These look out of order, I'm seeing that around the files here in general

libiocage/lib/Jails.py
```diff
@@ -26,122 +28,84 @@ def __init__(self,
...
-
-        jails = self._get_existing_jails()
-
-        if filters is not None:
-            return self._filter_jails(jails, filters)
-        else:
-            return jails
+        self._filters = None
+        self.filters = filters
+        list.__init__(self, [])
```
skarekrow: Debug?

libiocage/cli/__init__.py
```diff
@@ -95,11 +142,16 @@ def get_command(self, ctx, name):
...
             return mod.cli
         except (ImportError, AttributeError):
+            raise
             return
 
 
+@click.option("--log-level", "-d", default=None)
 @click.command(cls=IOCageCLI)
-@click.version_option(version="0.9.10 07/30/2017", prog_name="iocage",
+@click.version_option(version="0.2.11 08/29/2017", prog_name="ioc",
```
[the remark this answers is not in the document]
gronke: I started with 0.0.1 and everytime it will introduce new features I'd love to bump the minor version. For breaking changes (after 1.0.0) the major version number will increase. We should discuss and document the versioning system!

libiocage/cli/list.py
```diff
@@ -24,11 +24,15 @@
...
 import click
 import texttable
+import typing
 
 import libiocage.lib.Host
 import libiocage.lib.Jails
+import libiocage.lib.JailFilter
 import libiocage.lib.Logger
 
+supported_output_formats = ['table', 'csv', 'list']
```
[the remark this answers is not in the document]
gronke: Maybe now. That's super easy. But *csv* and *list* come with some nice optimizations - they stream their output. For JSON that would be useless :) But I will add it because it's a low hanging fruit!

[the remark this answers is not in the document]
skarekrow: `yaml` is even lower hanging fruit! It should also buy us XML iirc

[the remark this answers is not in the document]
igalic: no no no double no

gronke:
@skarekrow we can work on the license issue in a different PR :)

libiocage/cli/start.py
gronke: Example pls :)

libiocage/cli/list.py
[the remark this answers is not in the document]
gronke: @igalic I can understand that you want XML so badly, but we can't just disable all other list options just for that! :trollface: @skarekrow would be very disappointed if we drop the table output!

libiocage/cli/start.py
gronke: No exit! Was a left-over from the previous implementation. Will be cleaned with the next commit!

libiocage/lib/Jail.py
gronke: I was following different approaches to wrap the generators, because I think it's important that libiocage can work in both modes. Of course I'm trying to keep the overhead caused by wrapping as low as possible. Expect this code to be removed before cleaning up for a merge!

libiocage/lib/Jails.py
gronke: Nope. Line 31 makes the `_filters` property available, the line below calls the `filters` property setter.

libiocage/lib/Jails.py
gronke: Didn't care about the order yet. Let's put in a static analyzer rule in addition to the contribution guidelines.

I suggest to tackle those kind of changes in a separate PR, so that we concentrate on a specific feature in a PR. Too often features have implications to other parts of libiocage as well. But I'm pretty sure this will change in the near future after the interfaces are well tested.

https://github.com/iocage/libiocage/issues/47 is the ticket for that

libiocage/cli/start.py
skarekrow:
```
exit(1) if len(failed_jails) > 0 else exit(0)
```

gronke: Got your point now!! That's even a bit shorter:

```python
exit(1*bool(len(failed_jails)))
```

skarekrow: Readability counts in python though ;) It's not always about shortest in that sense.

libiocage/lib/Jail.py
```diff
@@ -661,22 +711,21 @@ def _teardown_mounts(self):
                 )
 
     def _resolve_name(self, text):
+
+        if (text is None) or (len(text)==0):
```
skarekrow: I know, you know I hate parenthesis, but I think this is a good example. It makes de-compacting that hard on my eyes. Instead of `if test is None or len(text) == 0`, it's easier to see those elements to me. I won't mark as needs changes, because it's your personal style ;)

But just felt I had to say it atleast! :D

libiocage/lib/Logger.py
```diff
@@ -85,7 +85,7 @@ def __init__(self, print_level=None, log_directory="/var/log/iocage"):
 
     @property
     def default_print_level(self):
-        return "spam"
+        return "info"
```
skarekrow: info now? You told me that printed, does it not now?

libiocage/lib/JailConfig.py
```diff
@@ -213,7 +213,7 @@ def _set_name(self, name, **kwargs):
             self["id"] = name
         else:
             try:
-                self["id"] = str(uuid.UUID(name)) # legacy support
+                self["id"] = str(uuid.UUID(name))  # legacy support
```
skarekrow: Since this is for legacy support, I think `uuid` would be a better indicator of what the id is. As id can be a bit ambiguous

gronke: The `ID` brings together all the concepts of name and uuid :) It's the only true identifier of a jail.

libiocage/lib/JailFilter.py
```diff
@@ -0,0 +1,165 @@
...
+        """
+        Returns True if the value matches the term
+        """
+        for filter_value in self:
+
+            if self.key == "name":
+                if self._validate_name_filter_string(filter_value) is False:
+                    raise libiocage.lib.errors.JailFilterInvalidName(
+                        filter_value,
+                        logger=self.logger
```
igalic: The class Term has no attribute `logger`, if it needs one, we should extend its `__init__`

igalic: there's now a patch for this #62

libiocage/cli/list.py
```diff
@@ -23,12 +23,17 @@
...
+import json
 import texttable
+import typing
 
 import libiocage.lib.Host
 import libiocage.lib.Jails
+import libiocage.lib.JailFilter
 import libiocage.lib.Logger
 
+supported_output_formats = ['table', 'csv', 'list', 'json']
```
xaque208: Yay json!

PR #46, opened 2017-08-30
Add EOL tags to the fetched list output

skarekrow:
This replicates iocage's current output
```
╰─ ioc fetch
[0] 10.1-RELEASE (EOL)
[1] 10.2-RELEASE (EOL)
[2] 10.3-RELEASE
[3] 10.4-BETA2
[4] 11.0-RELEASE
[5] 11.1-RELEASE
[6] 9.3-RELEASE (EOL)

Type the number of the desired RELEASE
Press [Enter] to fetch the default selection (11.1-RELEASE) [5]:
```

### DIFF
--- a/libiocage/lib/Distribution.py
+++ b/libiocage/lib/Distribution.py
@@ -2,6 +2,9 @@ import os
 import platform
 import re
 import urllib.request
+from typing import List
+
+import requests
 
 import libiocage.lib.Release
 import libiocage.lib.errors
@@ -63,6 +66,7 @@ class Distribution:
         response = resource.read().decode(charset if charset else "UTF-8")
 
         found_releases = self._parse_links(response)
+        eol_list = self._get_eol_list()
 
         available_releases = sorted(
             map(  # map long HardenedBSD release names
@@ -79,7 +83,8 @@ class Distribution:
                 name=x,
                 host=self.host,
                 zfs=self.zfs,
-                logger=self.logger
+                logger=self.logger,
+                eol=self._check_eol(x, eol_list)
             ),
             available_releases
         ))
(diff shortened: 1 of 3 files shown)


libiocage/lib/Distribution.py
```diff
@@ -97,6 +101,41 @@ def _filter_available_releases(self, release_name):
...
+
+        return self._get_eol_releases(releases, eol_releases)
+
+    def _get_eol_releases(self, releases: List[str], eol: List[str]) -> \
+            List[str]:
+        releases_eol = []
+        if self.host.distribution.name == "FreeBSD":
+            for rel in releases:
+                if rel in eol:
+                    rel = f"{rel} (EOL)"
```
gronke: I suggest to add a property to Release that is set to True when the Release is EOL. That way we don't need to change the internal name. The Release name is involved in calculating the dataset location and much more.

When rendering you can just check

```python
release_name = release.name
if release.eol is True:
  printed_name += " (EOL)"
```

Or even another property on Release that gives you a releases name with annotations for printing:

```
@property
def annotated_name(self):
  annotations = set()
  if self.eol is True:
    annotations.add("EOL")
  if len(annotations) > 0:
    return "{} ({})".format(release.name, ", ".join(annotations))
  return f"{self.name} {suffix}"
```

By using the latter, we could make sure `(EOL)` gets even printed when listing existing jails, if we want to.

libiocage/lib/Distribution.py
```diff
@@ -97,6 +101,41 @@ def _filter_available_releases(self, release_name):
         arch = release_name.split("-")[-2:][0]
         return self.host.processor == arch
 
+    def _get_eol_list(self, releases: List[str]) -> List[str]:
+        """Scrapes the FreeBSD website and returns a list of EOL RELEASES"""
+        _eol = "https://www.freebsd.org/security/unsupported.html"
+        req = requests.get(_eol)
```
gronke: Can we cache this in a file next to `default.json` and check if atime is not older than 3 days (or another reasonable time)? Maybe `eol.json`?

libiocage/lib/Distribution.py
skarekrow: Great suggestion!

libiocage/lib/Distribution.py
skarekrow: That may not work, considering users can disable atime altogether. We should think of a less fragile mechanic. Perhaps we encode the date in the JSON? Check that for 3 days later?

igalic: we can always write the etag and/or other cache identifying headers into eol.json

skarekrow:
The suggestion has been merged, minus the `eol.json`.

libiocage/lib/Distribution.py
skarekrow: We could also just use https://pypi.python.org/pypi/CacheControl/0.6 and bypass having a file at all still.

PR #54, opened 2017-09-01
implement type-hints for all (some) our functions

igalic:
gradually add types on all functions and variables.
this effort is guided by mypy.


libiocage/lib/Distribution.py
```diff
@@ -27,7 +27,7 @@ class DistributionGenerator:
     def __init__(self, host, zfs=None, logger=None):
         self.logger = libiocage.lib.helpers.init_logger(logger)
         self.zfs = libiocage.lib.helpers.init_zfs(zfs)
-        libiocage.lib.helpers.init_host(self, host)
+        self.host = libiocage.lib.helpers.init_host(host, self.logger)
```
gronke: I prefer to use named arguments - at least for logger and those kind of dependencies. It's much easier to recognize broken interfaces when being explicit.

igalic: ack, will change

libiocage/lib/events.py
```diff
@@ -15,7 +17,7 @@ class IocageEvent:
     Base class for all other iocage events
     """
 
-    HISTORY = []
+    HISTORY: List[Any] = []
```
igalic: this is not ideal, but i have idea how to tell it the type of this class

libiocage/lib/helpers.py
```diff
@@ -1,38 +1,31 @@
...
+        return get_zfs()
 
 
 def get_zfs():
     return libzfs.ZFS(history=True, history_prefix="<iocage>")
 
 
-def init_host(self, host=None):
+def init_host(host: libiocage.lib.Host.Host=None, logger:
+              libiocage.lib.Logger.Logger=None) -> libiocage.lib.Host.Host:
```
igalic: this return type is now incompatible

libiocage/lib/helpers.py
```diff
@@ -1,38 +1,31 @@
...
-            logger = self.logger
-        except:
-            logger = None
+        return host
 
-        try:
-            self.host = self._class_host(logger=logger)
-        except:
-            self.host = libiocage.lib.Host.HostGenerator(logger=logger)
+    return libiocage.lib.Host.HostGenerator(logger=logger)
```
igalic: with this return

libiocage/lib/events.py
gronke: Guess I can explain that: `List[libiocage.lib.events.IocageEvent]`

The class attribute, not the instance attribute contains a list of all events occurred. This way every event has access to the full history.

libiocage/lib/helpers.py
gronke: We're returning a HostGenerator. Internally we don't use the wrappers that flatten the generator to a result list.

libiocage/lib/Distribution.py
```diff
@@ -25,12 +25,11 @@ class DistributionGenerator:
     mirror_link_pattern = r"a href=\"([A-z0-9\-_\.]+)/\""
 
     def __init__(self, host, zfs=None, logger=None):
-        libiocage.lib.helpers.init_logger(self, logger)
-        libiocage.lib.helpers.init_zfs(self, zfs)
-        libiocage.lib.helpers.init_host(self, host)
+        self.logger = libiocage.lib.helpers.init_logger(logger)
+        self.zfs = libiocage.lib.helpers.init_zfs(zfs)
+        self.host = libiocage.lib.helpers.init_host(host, self.logger)
```
gronke: Please use named attributes for everything, but the first argument.

libiocage/lib/Jails.py
```diff
@@ -23,9 +23,11 @@ def __init__(self,
                  logger=None,
                  zfs=None):
 
-        libiocage.lib.helpers.init_logger(self, logger)
-        libiocage.lib.helpers.init_zfs(self, zfs)
-        libiocage.lib.helpers.init_host(self, host)
+        self.logger = libiocage.lib.helpers.init_logger(logger)
+        self.host = libiocage.lib.helpers.init_host(host, self.logger)
+
+        # this could be wrong..er.. now?:
```
gronke: What's this comment for? Do you suspect something to be wrong with the following lines?

libiocage/lib/errors.py
```diff
@@ -388,7 +388,8 @@ def __init__(self, *args, **kwargs):
 
 class MissingFeature(IocageException, NotImplementedError):
 
-    def __init__(self, feature_name: str, plural: bool=False, *args, **kwargs):
+    def __init__(self, feature_name: str, plural: bool=False,
+                 *args, **kwargs) -> None:
```
gronke: Please break every argument into a new line. Like such:

```python
def __init__(
    self
    feature_name: str
    plural: bool=False,
    *args
    **kwargs
) -> None:
    pass
```

libiocage/lib/helpers.py
gronke: `HostGenerator` should be the type returned.

igalic: ack, thanks

libiocage/lib/Jails.py
igalic: so, we init_zfs(), and then we run the equivalent of the get_zfs() function, and if that's meant to be happening consistently, we may wanna just always do that in init_zfs()

libiocage/lib/errors.py
igalic: yapf doesn't work for python 3.6, and standard indent can't seem to be able to read minds 😛

gronke: I'm running this syntax on python3.6. Why should it not work?

libiocage/lib/events.py
igalic: python (or mypy) can't find the type `libiocage.lib.events.IocageEvent`

gronke: Within this file it's just `IocageEvent`, not the fqdn :trollface:

igalic: yeah, that's what I thought too, but that doesn't work either.

libiocage/lib/errors.py
igalic: done.

libiocage/lib/Distribution.py
```diff
@@ -36,7 +36,7 @@
 
 class DistributionGenerator:
 
-    _class_release = libiocage.lib.Release.ReleaseGenerator
+    _class_release = 'libiocage.lib.Release.ReleaseGenerator'
```
gronke: This should not be a string. Later on we instantiate this class.

libiocage/lib/Host.py
```diff
@@ -91,4 +89,4 @@ def processor(self):
 
 class Host(HostGenerator):
 
-    class_distribution = libiocage.lib.Distribution.DistributionGenerator
+    class_distribution = 'libiocage.lib.Distribution.DistributionGenerator'
```
gronke: it should have been `libiocage.lib.Distribution.Distribution` here. Line 35 should not be deleted as well.

libiocage/lib/Host.py
```diff
@@ -32,18 +32,16 @@
...
-        libiocage.lib.helpers.init_zfs(self, zfs)
+        self.logger = libiocage.lib.helpers.init_logger(logger)
+        self.zfs = libiocage.lib.helpers.init_zfs(zfs)
         self.datasets = libiocage.lib.Datasets.Datasets(
             root=root_dataset,
             logger=self.logger,
             zfs=self.zfs
         )
-        self.distribution = self._class_distribution(
+        self.distribution = libiocage.lib.Distribution.DistributionGenerator(
```
gronke: It is important to instantiate from the class attribute _class_distribution. If you use a non-generator class like Host, you don't expect a generator distribution.

libiocage/lib/JailConfig.py
```diff
@@ -79,8 +79,7 @@ def __init__(self,
 
         dict.__init__(self)
 
-        libiocage.lib.helpers.init_logger(self, logger)
-
+        self.logger = libiocage.lib.helpers.init_logger(logger)
```
gronke: I like to use the output from the helper to store explicitly as self.logger, but we should remain passing self as first attribute. This way we do not need to pass other named arguments explicitly later on.

libiocage/lib/Jails.py
```diff
@@ -46,9 +46,11 @@ def __init__(self,
                  logger=None,
                  zfs=None):
 
-        libiocage.lib.helpers.init_logger(self, logger)
-        libiocage.lib.helpers.init_zfs(self, zfs)
-        libiocage.lib.helpers.init_host(self, host)
+        self.logger = libiocage.lib.helpers.init_logger(logger)
+        self.host = libiocage.lib.helpers.init_host(host, logger=self.logger)
```
gronke: Here's one example why we should pass self as first argument to the init_ helpers.

libiocage/lib/errors.py
```diff
@@ -412,7 +412,11 @@ def __init__(self, *args, **kwargs):
 
 class MissingFeature(IocageException, NotImplementedError):
 
-    def __init__(self, feature_name: str, plural: bool=False, *args, **kwargs):
+    def __init__(self,
+                 feature_name: str,
+                 plural: bool=False,
+                 *args,
+                 **kwargs) -> None:
```
gronke: Please use this style for multi-line method interfaces
```python
def __init__(self,
    feature_name: str,
    plural: bool=False,
    *args,
    **kwargs
) -> None:
```

libiocage/lib/Host.py
gronke: @igalic as we discussed offline, this change should be reverted

libiocage/lib/Host.py
igalic: n.b.: In `class HostGenerator` this property was `_class_distribution` (yay ☭) but here it's without the leading `_`.

libiocage/lib/Host.py
```diff
@@ -91,4 +91,4 @@ def processor(self):
 
 class Host(HostGenerator):
 
-    class_distribution = libiocage.lib.Distribution.DistributionGenerator
+    _class_distribution = libiocage.lib.Distribution.DistributionGenerator
```
igalic: is this supposed to be a generator?
is it supposed to be `class_distribution` and not `_class_distribution`

gronke: As rule of thumb: when the class is not a generator it is not supposed to expose or use generators at all. So, this should be a Distribution, not DistributionGenerator.

gronke:
@igalic besides that `init_*` functions should remain having `self` as first argument, so that we don't need to pass other attributes (like logger, host, etc) on every helpers.init_* call. It's easier to update on inter-dependency changes in future updates as well.

PR #55, opened 2017-09-01
Feature: Rename Jails

### LINKED ISSUE #52: Feature: Rename jail
A jails name also defines the jails dataset location relative to the activated zpool. That's why it's important to do more than just renaming the property. Certainly we must force a jail to be stopped for renaming it.

### Requirements
- When renaming a Jail
  - [ ] an exception is raised when the jail is started
  - [ ] moves the dataset immediately
  - [ ] changes the `host_hostname` when it was manually set (available in config.data) and matches the old jail name

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/48866814-feature-rename-jail?utm_campaign=plugin&utm_content=tracker%2F70027116&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F70027116&utm_medium=issues&utm_source=github).
</bountysource-plugin>

gronke:
closes #52

### Changes
- Introduced ioc command `rename`
- Renames a dataset if the jail was stopped and the new name is valid

### Usage
```
ioc rename my-jail new-jail-name
```

### DIFF
--- a/libiocage/cli/rename.py
+++ b/libiocage/cli/rename.py
@@ -1,0 +1,46 @@
+# Copyright (c) 2014-2017, iocage
+# All rights reserved.
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted providing that the following conditions
+# are met:
+# 1. Redistributions of source code must retain the above copyright
+#    notice, this list of conditions and the following disclaimer.
+# 2. Redistributions in binary form must reproduce the above copyright
+#    notice, this list of conditions and the following disclaimer in the
+#    documentation and/or other materials provided with the distribution.
+#
+# THIS SOFTWARE IS PROVIDED BY THE AUTHOR ``AS IS'' AND ANY EXPRESS OR
+# IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
+# WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+# ARE DISCLAIMED.  IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR ANY
+# DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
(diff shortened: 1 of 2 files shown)


libiocage/lib/Jail.py
```diff
@@ -289,6 +289,23 @@ def destroy(self, force=False):
 
         self.storage.delete_dataset_recursive(self.dataset)
 
+    def rename(self, new_name: str):
```
igalic: please add `-> None` as return type

libiocage/lib/Jail.py
```diff
@@ -289,6 +289,23 @@ def destroy(self, force=False):
...
+        """
+        Change the name of a jail
+        """
+
+        self.require_jail_existing()
+        self.require_jail_stopped()
+
+        current_id = self.config["id"]
+        dataset = self.dataset
+        self.config["name"] = new_name  # validates new_name
```
igalic: huh, how??

libiocage/cli/rename.py
```diff
@@ -0,0 +1,46 @@
...
+def cli(ctx, jail, name):
+    """
+    Rename a stopped jail.
+    """
+
+    logger = ctx.parent.logger
+    try :
+        ioc_jail = libiocage.lib.Jail.Jail(jail, logger=logger)
+        ioc_jail.rename(name)
+    except libiocage.lib.errors.IocageException:
```
igalic: what kind of output will this produce?

igalic:
unrelated to this, travis is unhappy:

```
+./libiocage/lib/JailConfigResolver.py:86:18: F821 undefined name 'isintance'
+./libiocage/lib/JailConfig.py:617:13: F841 local variable 'no_user_defined_property' is assigned to but never used
```

we should fix this first before merging anything else

libiocage/lib/Jail.py
skarekrow: It should throw a KeyError I believe, but if so it's missing a try block.

PR #66, opened 2017-09-03
Fix inconsistencies in fetch and create

igalic:
This patchset is the product of a debugging session with @gronke, in
which we fixed the fetching (and extraction!) of releases.

Speeding it up by orders of magnitude, by using rsync. It does a much
better job of preserving hardlinks, rather than duplicating the same
file 140 times.

We've also fixed the handling of the release - it can now be None, which
will eventually allow us to create EMPTY jails.

On the other end, Jail.create() now will never create a jail named None.

### DIFF
--- a/libiocage/lib/Jail.py
+++ b/libiocage/lib/Jail.py
@@ -390,7 +390,7 @@ class JailGenerator:
 
         backend = None
 
-        is_basejail = self.config["basjail"] == True
+        is_basejail = self.config.get("basejail", False)
         if not is_basejail:
             backend = libiocage.lib.StandaloneJailStorage.StandaloneJailStorage
         if is_basejail and self.config["basejail_type"] == "nullfs":
--- a/libiocage/lib/Jail.py
+++ b/libiocage/lib/Jail.py
@@ -350,6 +350,9 @@ class JailGenerator:
                 The jail is created from the release matching the name provided
         """
 
+        if self.config["id"] is None:
+            self.config["name"] = str(uuid.uuid4())
+
         self.require_jail_not_existing()
 
         # check if release exists
@@ -372,9 +375,6 @@ class JailGenerator:
 
         release = filteres_released[0]
         self.config["release"] = release.name
-
-        if not self.config["id"]:
-            self.config["name"] = str(uuid.uuid4())
 
         self.logger.verbose(
             f"Creating jail '{self.config['id']}'",
--- a/libiocage/lib/Jail.py
+++ b/libiocage/lib/Jail.py
@@ -390,7 +390,7 @@ class JailGenerator:
 
         backend = None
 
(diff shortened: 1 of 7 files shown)


libiocage/lib/Jail.py
```diff
@@ -390,7 +390,7 @@ def create(self, release_name):
 
         backend = None
 
-        is_basejail = self.config["type"] == "basejail"
+        is_basejail = self.config["basejail"] == True
```
skarekrow: This should be `is_basejail = self.config.get("basejail", True)` as that's pythonic and much nicer to read.

skarekrow:
It would be better if we can use a library for rsync as this being a new dependency, it would be great not to shell out for.

igalic:
it would, but this quick fixing shellout  already makes us 10-100 times faster.

libiocage/lib/Jail.py
igalic: i did is True, flake8 could've done a better job suggesting what you suggested :D

skarekrow:
Ok we can explore that later :D great idea with rsync. I approve @gronke, damn phone prevents it

libiocage/lib/Jail.py
igalic: n.b.: in other places we already use

```python
    @property
    def basejail_backend(self):

        if self.config["basejail"] is False:
            return None

        if self.config["basejail_type"] == "nullfs":
            return libiocage.lib.NullFSBasejailStorage.NullFSBasejailStorage

        if self.config["basejail_type"] == "zfs":
            return libiocage.lib.ZFSBasejailStorage.ZFSBasejailStorage

        return None
```

igalic:
@skarekrow i don't see what exactly you're asking me to change, but if it's the `== True` thing, see: https://github.com/iocage/libiocage/pull/66#discussion_r136728544

skarekrow:
@igalic those are just truthy tests, not the same thing. You're doing an assignment and that syntax is odd for that. What I supplied is the better alternative.

igalic:
@skarekrow updated

libiocage/lib/Jail.py
```diff
@@ -390,7 +390,7 @@ def create(self, release_name):
 
         backend = None
 
-        is_basejail = self.config["type"] == "basejail"
+        is_basejail = self.config.get("basejail", False)
```
gronke: self.config["basejail"] is already a boolean because it returns the parsed user input.

skarekrow: Sure. It grabs a value from a dictionary key. If it doesn't exist it becomes the default value. On this case False. 

The equivalent of try;except KeyError;foo 

>

gronke: This is so smart! Thanks for the clarification.

PR #73, opened 2017-09-05
Python tracing for better performance

gronke:
The environment variable `IOCAGE_TRACE` can point to a directory where iocage will store simple trace logs counting the number of function calls.

```sh
setenv IOCAGE_TRACE=/tmp/iocage-trace
python3.6 . list
cat /tmp/iocage-trace/*
```

Looking at this traces will improve our performance in the long run 🏅

### DIFF
--- a/__main__.py
+++ b/__main__.py
@@ -33,5 +33,32 @@ import click
 
 from libiocage.cli import cli
 
+
+def main_safe():
+  try:
+    main()
+  except BaseException as e:
+    return e
+
+
+def main():
+  cli(prog_name="iocage")
+
+
 if __name__ == "__main__":
-    cli(prog_name="iocage")
+
+  debug = False
+
+  try:
+    coverdir = os.environ["IOCAGE_TRACE"]
+    import trace
+    tracer = trace.Trace(
+      ignoredirs=[sys.prefix, sys.exec_prefix],
+      trace=0,
+      count=1)
+    tracer.run("main_safe()")
+    r = tracer.results()
+    r.write_results(show_missing=True, coverdir=coverdir)
+    print(f"Iocage Trace written to: {coverdir}")
+  except:
+    main()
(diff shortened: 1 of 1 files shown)


__main__.py
```diff
@@ -33,5 +33,32 @@
...
+
+
 if __name__ == "__main__":
-    cli(prog_name="iocage")
+
+  debug = False
+
+  try:
+    coverdir = os.environ["IOCAGE_TRACE"]
+    import trace
```
skarekrow: This should be imported at the top.

__main__.py
```diff
@@ -33,5 +33,32 @@
...
+  cli(prog_name="iocage")
+
+
 if __name__ == "__main__":
-    cli(prog_name="iocage")
+
+  debug = False
+
+  try:
+    coverdir = os.environ["IOCAGE_TRACE"]
```
skarekrow: This should be `coverdir = os.environ.get("IOCAGE_TRACE", DEFAULTVALUE)`, otherwise coverdir will be `None` and may cause issues.

__main__.py
```diff
@@ -33,5 +33,32 @@
...
+    import trace
+    tracer = trace.Trace(
+      ignoredirs=[sys.prefix, sys.exec_prefix],
+      trace=0,
+      count=1)
+    tracer.run("main_safe()")
+    r = tracer.results()
+    r.write_results(show_missing=True, coverdir=coverdir)
+    print(f"Iocage Trace written to: {coverdir}")
+  except:
```
skarekrow: What exceptions are we looking for? Having a broad `except` clause may catch things we don't necessarily intend to catch. We should be as explicit as possible with those.

__main__.py
gronke: We only need it for debugging, so I want to only include it when it's required.

__main__.py
gronke: Will be removed entirely by addressing https://github.com/iocage/libiocage/pull/73#discussion_r137129016

PR #80, opened 2017-09-10
Feature: Resource Limits

### LINKED ISSUE #21: Feature: Resource Limits
FreeBSD allows to define resource limits for jails, as seen in the documentation:
https://www.freebsd.org/doc/handbook/security-resourcelimits.html

See also:
- [rctl man page](https://www.freebsd.org/cgi/man.cgi?query=rctl&sektion=8&apropos=0&manpath=FreeBSD+11-current)
- [rctl.conf man page](https://www.freebsd.org/cgi/man.cgi?query=rctl.conf&apropos=0&sektion=5&manpath=FreeBSD+11-current&format=html)
- [expand_number man page](https://www.freebsd.org/cgi/man.cgi?query=expand_number&sektion=3&apropos=0&manpath=FreeBSD+11-current)

## Requirements
- [ ] Resource limits can be defined via JailConfig
- [ ] Memory can be limited via `limit_memoryuse` option
- When limiting memory
  - [ ] a jail may be running
  - [ ] values can be provided in human readable format according to expand_number 
- [ ] CPU usage on a single CPU core in percent can be limited via `limit_cputime` option
- [ ] Other limits can be set via `limit_<RESOURCE>` according to the rctl man page RESOURCES section

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/48592986-feature-resource-limits?utm_campaign=plugin&utm_content=tracker%2F70027116&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F70027116&utm_medium=issues&utm_source=github).
</bountysource-plugin>

gronke:
closes #21

### DIFF
--- a/libiocage/lib/Jail.py
+++ b/libiocage/lib/Jail.py
@@ -192,6 +192,7 @@ class JailGenerator(libiocage.lib.Resource.JailResource):
             self._configure_routes()
             yield jailVnetConfigurationEvent.end()
 
+        self._limit_resources()
         self._configure_nameserver()
 
         if self.config["jail_zfs"] is True:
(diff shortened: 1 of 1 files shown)


libiocage/lib/Jail.py
```diff
@@ -664,6 +665,50 @@ def _stop_vimage_network(self) -> None:
...
+            try:
+                limit, action = self.config[resource].split(":", maxsplit=1)
+                command = [
+                    "/usr/bin/rctl",
+                    "-a",
+                    f"jail:{self.identifier}:{resource}:{action}={limit}"
+                ]
+                libiocage.lib.helpers.exec(command, logger=self.logger)
+            except KeyError:
+                pass
```
igalic: i don't understand why we're not throwing, ir at least logging an error if a resource limit is (in our code) considered invalid

gronke: updated issue #142 to cover this idea

PR #82, opened 2017-09-11
Feature: Hook script environment variables

### LINKED ISSUE #70: Feature: support for resolving runtime variables [$50]
As discussed through IRC adding option to use variables on specific attributes that are resolved on runtime would make using exec.pre/post_start/stop a lot easier. Native jail with jail.conf supports this out of the box ([jail.conf](https://www.freebsd.org/cgi/man.cgi?query=jail.conf&apropos=0&sektion=5&manpath=FreeBSD+11.1-RELEASE+and+Ports&arch=default&format=html), but I think it's not possible via cmd line using `jail -c ...` 

A workaround I did on legacy jail was to add jail name while calling exec scripts https://github.com/urosgruber/iocage/commit/ca766a772d1468b32d10a99888efe49a748e72e3

Example of an old script we used post_start script that is using jail name as an argument but could  be improved by providing JID or some other parameter 

```
#!/bin/sh

# get current jid
_jid=`jls -j $1 jid 2>/dev/null`
_ip=`jls -j $1 ip4.addr 2>/dev/null`
_name=$1

# cpuset
# cpuset -c -l 0 -j ${_jid}

# mem limit
rctl -a jail:$1:memoryuse:deny=256M

# pf
( pfctl -a jails -sr ; echo "pass on lo0 from ${_ip} to ${_ip}" ) | pfctl -a jails -f -

# update started time and status
mysql -ss -D db -e "UPDATE service SET started=NOW(), status='running', jid=${_jid} WHERE name='$1';" 2> /dev/null
```
Similarly for post_stop script and pre_ scripts.

Nice to have would be to specify a folder **pre_start.d**  where all executable files would be executed with same parameters. example: `ioc set exec_prestart="/my/prestart.d/ ${jid}"`

<bountysource-plugin>

---
There is a **[$50 open bounty](https://www.bountysource.com/issues/48978831-feature-support-for-resolving-runtime-variables?utm_campaign=plugin&utm_content=tracker%2F70027116&utm_medium=issues&utm_source=github)** on this issue. Add to the bounty at [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F70027116&utm_medium=issues&utm_source=github).
</bountysource-plugin>

gronke:
closes #70 

All Jail configuration properties (including defaults if not overridden) are passed as environment variables to the `exec_prestart`, `exec_start`, `exec_poststart` and `exec_prestop` commands.

### Sample of iocage environment variables

```
IOCAGE_DEVFS_RULESET=4
IOCAGE_ALLOW_MOUNT_TMPFS=0
IOCAGE_MOUNT_DEVFS=1
IOCAGE_ALLOW_SOCKET_AF=0
IOCAGE_EXEC_START=/bin/sh /etc/rc
IOCAGE_CHILDREN_MAX=0
IOCAGE_EXEC_STOP=/bin/sh /etc/rc.shutdown
IOCAGE_RESOLVER=/etc/resolv.conf
IOCAGE_EXEC_POSTSTART=/tmp/foo.sh
IOCAGE_ALLOW_MOUNT_ZFS=0
IOCAGE_EXEC_TIMEOUT=60
IOCAGE_CLONEJAIL=yes
IOCAGE_EXEC_PRESTART=/usr/bin/true
IOCAGE_DEFAULTROUTER6=-
IOCAGE_ID=foo2
IOCAGE_EXEC_CLEAN=1
IOCAGE_JAIL_ZFS=no
IOCAGE_ALLOW_CHFLAGS=0
IOCAGE_EXEC_FIB=1
IOCAGE_ENFORCE_STATFS=2
IOCAGE_VNET=no
IOCAGE_DEFAULTROUTER=-
IOCAGE_ALLOW_MOUNT_DEVFS=0
IOCAGE_MOUNT_FDESCFS=1
IOCAGE_NAME=foo2
IOCAGE_IP4=new
IOCAGE_BASEJAIL=yes
IOCAGE_IP4_SADDRSEL=1
IOCAGE_HOST_DOMAINNAME=None
IOCAGE_SYSVMSG=new
IOCAGE_IP6=new
IOCAGE_SYSVSEM=new
IOCAGE_SECURELEVEL=2
IOCAGE_PRIORITY=0
IOCAGE_ALLOW_MOUNT_NULLFS=0
IOCAGE_EXEC_POSTSTOP=/usr/bin/true
IOCAGE_EXEC_PRESTOP=/usr/bin/true
IOCAGE_SYSVSHM=new
IOCAGE_ALLOW_MOUNT_PROCFS=0
IOCAGE_TAGS=-
IOCAGE_IP6_SADDRSEL=1
IOCAGE_ALLOW_SYSVIPC=0
IOCAGE_ALLOW_MOUNT=0
IOCAGE_RELEASE=11.1-RELEASE
IOCAGE_ALLOW_RAW_SOCKETS=0
IOCAGE_MAC_PREFIX=02ff60
IOCAGE_ALLOW_QUOTAS=0
IOCAGE_JAIL_PATH=/iocage/jails/foo2/root
IOCAGE_LEGACY=None
IOCAGE_BOOT=no
IOCAGE_ALLOW_SET_HOSTNAME=1
IOCAGE_STOP_TIMEOUT=30
```

### DIFF
--- a/libiocage/lib/Jail.py
+++ b/libiocage/lib/Jail.py
@@ -1023,6 +1023,8 @@ class JailGenerator(JailResource):
             prop_name = f"IOCAGE_{prop.upper()}"
             jail_env[prop_name] = self.getstring(prop)
 
+        jail_env["IOCAGE_JAIL_PATH"] = self.root_dataset.mountpoint
+
         return jail_env
 
     @property
--- a/libiocage/lib/Jail.py
+++ b/libiocage/lib/Jail.py
@@ -275,6 +275,7 @@ class JailGenerator(JailResource):
         if quick is False:
             self._save_autoconfig()
 
+        self._run_hook("prestart")
         self._launch_jail()
 
         yield jailLaunchEvent.end()
@@ -300,6 +301,8 @@ class JailGenerator(JailResource):
             self._start_services()
             yield jailServicesStartEvent.end()
 
+        self._run_hook("poststart")
+
     @property
     def basejail_backend(self):
 
@@ -313,6 +316,24 @@ class JailGenerator(JailResource):
             return libiocage.lib.ZFSBasejailStorage.ZFSBasejailStorage
 
         return None
+
+    def _run_hook(self, hook_name: str):
+
+        key = f"exec_{hook_name}"
+        value = self.config[key]
+
+        if value == "/usr/bin/true":
+            return
+
+        self.logger.verbose(
+            f"Running {hook_name} hook for {self.humanreadable_name}"
+        )
+
+        return libiocage.lib.helpers.exec(
+            value.split(" "),
+            logger=self.logger,
+            env=self.env
+        )
 
     def _start_services(self):
         command = self.config["exec_start"].strip().split()
@@ -342,6 +363,8 @@ class JailGenerator(JailResource):
         jailDestroyEvent = events.JailDestroy(self)
         jailNetworkTeardownEvent = events.JailNetworkTeardown(self)
         jailMountTeardownEvent = events.JailMountTeardown(self)
+
+        self._run_hook("prestop")
 
         yield jailDestroyEvent.begin()
         self._destroy_jail()
@@ -399,6 +422,12 @@ class JailGenerator(JailResource):
     def _force_stop(self):
 
         successful = True
+
+        try:
+            self._run_hook("prestop")
+        except:
+            successful = False
+            self.logger.warn("pre-stop script failed")
 
         try:
             self._destroy_jail()
@@ -523,7 +552,10 @@ class JailGenerator(JailResource):
         command = ["/usr/sbin/jexec", self.identifier] + command
 
         return libiocage.lib.helpers.exec(
-            command, logger=self.logger, **kwargs
+            command,
+            logger=self.logger,
+            env=self.env,
+            **kwargs
         )
 
     def passthru(self, command):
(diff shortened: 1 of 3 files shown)


libiocage/lib/Jail.py
```diff
@@ -314,6 +317,24 @@ def basejail_backend(self):
...
+
+        if value == "/usr/bin/true":
+            return
+
+        self.logger.verbose(
+            f"Running {hook_name} hook for {self.humanreadable_name}"
+        )
+
+        return libiocage.lib.helpers.exec(
+            value.split(" "),
```
skarekrow: So if we split on white space, you will lose the exec_start and exec_stop and such. Since it's /bin/sh/etc/rc*

libiocage/lib/Jail.py
```diff
@@ -401,6 +424,12 @@ def _force_stop(self):
         successful = True
 
         try:
+            self._run_hook("prestop")
+        except:
+            successful = False
```
skarekrow: I can't see in this diff where successful is used. Is it?

gronke: Yes, it is. Right at the end of the method

libiocage/lib/Jail.py
gronke: We need to set the subprocess argument shell=True or split the command and it's arguments into a list.

skarekrow: We shouldn't shell with subprocess. That's just another can of worms :D We should yes make it a list (why I marked this as needs changes) or we can use shlex to parse it into the list too.

PR #83, opened 2017-09-11
Feature: Tags

### LINKED ISSUE #75: Feature: Tags
In addition to a resource name, jails may have one or more tags that makes them selectable.

The general functionality to provide tags is embedded in libiocage, but we need to expose the feature via the command line interface to make it usable.

For compatibility with previous versions the tags should be stored as comma separated string. Another thing to consider is the previously used jail `tag` configuration option which should be seamlessly merged into the tags when it exists.

### Requirements
- [x] a jail can have one or more tags
- [x] a jails tags can be shown in list command output
- [x] jail filters can match individual tags
- [x] jail config property `tags` can be set from a comma separated string
- [x] jail config property `tags` can be set from a list of strings 
- When no user defined `tags` property was found in a jail's configuration
  - [x] setting a `tag` JailConfig item sets the deprecated `tag` property
  - [x] setting `tags` removed any existing `tag` property from the configuration
- When user defined `tags` are already existing
  - [x] setting a `tag` that already exists in `tags` makes it the first item in `tags` 
  - [x] setting a `tag` that does not exist in `tags` yet, adds it as first item

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/49121178-feature-tags?utm_campaign=plugin&utm_content=tracker%2F70027116&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F70027116&utm_medium=issues&utm_source=github).
</bountysource-plugin>

gronke:
closes #75 

## Requirements (copy)
- [x] a jail can have one or more tags
- [x] a jails tags can be shown in list command output
- [x] jail filters can match individual tags
- [x] jail config property `tags` can be set from a comma separated string
- [x] jail config property `tags` can be set from a list of strings 
- When no user defined `tags` property was found in a jail's configuration
  - [x] setting a `tag` JailConfig item sets the deprecated `tag` property
  - [x] setting `tags` removed any existing `tag` property from the configuration
- When user defined `tags` are already existing
  - [x] setting a `tag` that already exists in `tags` makes it the first item in `tags` 
  - [x] setting a `tag` that does not exist in `tags` yet, adds it as first item

### DIFF
--- a/libiocage/lib/Jail.py
+++ b/libiocage/lib/Jail.py
@@ -119,31 +119,13 @@ class JailResource(libiocage.lib.LaunchableResource.LaunchableResource):
     def dataset_name(self, value: str):
         self._dataset_name = value
 
-    def get(self, key):
+    def get(self, key: str) -> typing.Any:
         try:
             return self.jail.config[key]
-        except:
+        except KeyError:
             pass
 
-        try:
-            return self.resource.get(key)
-        except:
-            pass
-
-        return None
-
-    def getstring(self, key):
-        """
-        Returns a jail properties string or '-'
-
-        Args:
-            key (string):
-                Name of the jail property to return
-        """
-        return libiocage.lib.helpers.to_string(
-            self.get(key),
-            none="-"
-        )
+        return libiocage.lib.Resource.Resource.get(self, key)
 
 
 class JailGenerator(JailResource):
--- a/libiocage/lib/Jail.py
+++ b/libiocage/lib/Jail.py
@@ -119,6 +119,19 @@ class JailResource(libiocage.lib.LaunchableResource.LaunchableResource):
     def dataset_name(self, value: str):
         self._dataset_name = value
 
+    def get(self, key):
+        try:
+            return self.jail.config[key]
+        except:
+            pass
+
+        try:
+            return self.resource.get(key)
+        except:
+            pass
+
+        return None
+
     def getstring(self, key):
         """
         Returns a jail properties string or '-'
@@ -127,13 +140,10 @@ class JailResource(libiocage.lib.LaunchableResource.LaunchableResource):
             key (string):
                 Name of the jail property to return
         """
-
-        try:
-            return libiocage.lib.helpers.to_string(self.jail.config[key])
-        except:
-            pass
-
-        return libiocage.lib.Resource.Resource.getstring(self, key)
+        return libiocage.lib.helpers.to_string(
+            self.get(key),
+            none="-"
+        )
 
 
 class JailGenerator(JailResource):
--- a/libiocage/lib/JailConfig.py
+++ b/libiocage/lib/JailConfig.py
@@ -233,6 +233,37 @@ class JailConfig(dict, object):
     def _set_priority(self, value: typing.Union[int, str]):
         self.data["priority"] = str(value)
 
+    # legacy support
+    def _get_tag(self) -> str:
+
+        if self._has_legacy_tag is True:
+            return self.data["tag"]
+
+        try:
+            return self["tags"][0]
+        except KeyError:
+            return None
+
+    def _set_tag(self, value: str, **kwargs):
+
+        if (self._has_legacy_tag is True) or ("tags" not in self.data.keys()):
+            # store as deprecated `tag` for downwards compatibility
+            # setting `tags` overrides usage of the deprecated `tag` property
+            self.data["tag"] = value
+            return
+
+        tags = self["tags"]
+        if value in tags:
+            # remove the tag if it was existing
+            del tags[tags.index(value)]
+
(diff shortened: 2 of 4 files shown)


libiocage/lib/Jail.py
```diff
@@ -119,6 +119,19 @@ def dataset_name(self) -> str:
     def dataset_name(self, value: str):
         self._dataset_name = value
 
+    def get(self, key):
```
igalic: no type signature?

libiocage/lib/JailConfig.py
```diff
@@ -233,6 +233,37 @@ def _get_priority(self) -> int:
...
+
+        if self._has_legacy_tag is True:
+            return self.data["tag"]
+
+        try:
+            return self["tags"][0]
+        except KeyError:
+            return None
+
+    def _set_tag(self, value: str, **kwargs):
```
igalic: return type?

libiocage/lib/JailConfig.py
```diff
@@ -233,6 +233,37 @@ def _get_priority(self) -> int:
...
+        tags = self["tags"]
+        if value in tags:
+            # remove the tag if it was existing
+            del tags[tags.index(value)]
+
+        tags.insert(0, value)
+        self.data["tags"] = libiocage.lib.helpers.to_string(tags)
+
+    @property
+    def _has_legacy_tag(self):
```
igalic: return type?

libiocage/lib/Resource.py
```diff
@@ -264,6 +264,12 @@ def config_handler(self):
         handler = object.__getattribute__(self, f"config_{self.config_type}")
         return handler
 
+    def get(self, key):
```
igalic: type signature?

libiocage/lib/Resource.py
```diff
@@ -264,6 +264,12 @@ def config_handler(self):
...
+        try:
+            return self.__getattribute__(key)
+        except AttributeError:
+            return None
+
     def getstring(self, key):
```
igalic: type signature?

libiocage/lib/Resource.py
```diff
@@ -272,10 +278,10 @@ def getstring(self, key):
...
                 Name of the jail property to return
         """
-        try:
-            return libiocage.lib.helpers.to_string(self.__getattribute__(key))
-        except AttributeError:
-            return "-"
+        return libiocage.lib.helpers.to_string(
+            self.get(key),
+            none="-"
+        )
```
igalic: we've seen this exact code already above

igalic: jail is a resource
why can't it reuse this function?

libiocage/lib/Jail.py
```diff
@@ -119,6 +119,19 @@ def dataset_name(self) -> str:
...
+        try:
+            return self.jail.config[key]
+        except:
```
skarekrow: We should test for `KeyError`

libiocage/lib/Jail.py
```diff
@@ -119,6 +119,19 @@ def dataset_name(self) -> str:
...
+            pass
+
+        try:
+            return self.resource.get(key)
+        except:
```
skarekrow: We should test for `KeyError`

libiocage/lib/Jail.py
```diff
@@ -119,6 +119,19 @@ def dataset_name(self) -> str:
...
+            pass
+
+        return None
```
skarekrow: This is redundant. Functions always return `None` unless we say otherwise

libiocage/lib/Resource.py
gronke: The code here is now re-used instead of being re-implemented.

libiocage/lib/JailConfig.py
```diff
@@ -233,6 +233,37 @@ def _get_priority(self) -> int:
...
+    # legacy support
+    def _get_tag(self) -> str:
+
+        if self._has_legacy_tag is True:
+            return self.data["tag"]
+
+        try:
+            return self["tags"][0]
+        except KeyError:
+            return None
```
igalic: how is this even possible?

gronke: when not tag was set at all, tags is empty.. an if clause would have been the better fit

igalic: what i mean is, we declare `-> str`, but our return type is `Optional[str]`

gronke: Got you! This must be the case in other places as well. https://github.com/iocage/libiocage/issues/86 is the ticket for it.

PR #87, opened 2017-09-12
fix renaming

gronke:
fixes renaming jails

### DIFF
--- a/libiocage/cli/rename.py
+++ b/libiocage/cli/rename.py
@@ -39,7 +39,7 @@ def cli(ctx, jail, name):
     """
 
     logger = ctx.parent.logger
-    try :
+    try:
         ioc_jail = libiocage.lib.Jail.Jail(jail, logger=logger)
         ioc_jail.rename(name)
     except libiocage.lib.errors.IocageException:
--- a/libiocage/lib/Jail.py
+++ b/libiocage/lib/Jail.py
@@ -109,15 +109,20 @@ class JailResource(libiocage.lib.LaunchableResource.LaunchableResource):
         except:
             pass
 
+        return self._dataset_name_from_jail_name
+
+    @dataset_name.setter
+    def dataset_name(self, value: str) -> None:
+        self._dataset_name = value
+
+    @property
+    def _dataset_name_from_jail_name(self) -> str:
+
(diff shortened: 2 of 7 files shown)


libiocage/lib/JailConfig.py
```diff
@@ -37,6 +37,16 @@
 import libiocage.lib.Jail
 
 
+class SpecialProperties(dict):
```
igalic: this is name is a red flag :D

libiocage/cli/rename.py
```diff
@@ -39,7 +39,7 @@ def cli(ctx, jail, name):
     """
 
     logger = ctx.parent.logger
-    try :
+    try:
```
igalic: while you're touching these functions, you might as well add `-> None` to the the functions (and perhaps _other_ type signatures, too :)

libiocage/lib/Jail.py
```diff
@@ -409,11 +414,18 @@ def rename(self, new_name: str):
...
+            dataset.rename(new_dataset_name)
+            self.logger.verbose(
+                f"Dataset {current_dataset_name} renamed to {new_dataset_name}"
+            )
         except:
-            self.config["name"] = current_id
+            self.config["id"] = current_id
             raise
 
     def _force_stop(self):
```
igalic: `-> bool`

libiocage/lib/Jail.py
```diff
@@ -409,11 +414,18 @@ def rename(self, new_name: str):
...
         try:
-            dataset.rename(self.dataset_name)
+            new_dataset_name = self._dataset_name_from_jail_name
+            dataset.rename(new_dataset_name)
+            self.logger.verbose(
+                f"Dataset {current_dataset_name} renamed to {new_dataset_name}"
+            )
         except:
-            self.config["name"] = current_id
+            self.config["id"] = current_id
```
igalic: what's up with the moving back and forth between `id` and `name`

gronke: It's always ID. Name should always point to ID instead.

igalic: aye

gronke:
Closing in favor of #89

PR #88, opened 2017-09-14
Strict optional

igalic:
This PR is an attempt to correct type warnings issued by `mypy --strict-optional`
This catches a lot more issues where a variable was previously assigned `None`, or a return type should be `Optional`

Some of the following code is questionable and needs discussion & review.

### DIFF
--- a/libiocage/lib/Config/ConfigFile.py
+++ b/libiocage/lib/Config/ConfigFile.py
@@ -29,11 +29,11 @@ import libiocage.lib.Config.Prototype
 
 class ConfigFile(libiocage.lib.Config.Prototype.Prototype):
 
-    _file: str = None
+    _file: str
 
     def __init__(
         self,
-        file: str=None,
+        file: str,
         logger: 'libiocage.lib.Logger.Logger'=None
     ) -> None:
 
--- a/libiocage/lib/Host.py
+++ b/libiocage/lib/Host.py
@@ -63,7 +63,7 @@ class HostGenerator:
             host=self,
             logger=self.logger
         )
-        self._defaults = defaults
+        self._defaults = defaults if defaults is not None else {}
 
     @property
     def defaults(self) -> 'libiocage.lib.Resource.DefaultResource':
--- a/libiocage/lib/Host.py
+++ b/libiocage/lib/Host.py
@@ -40,8 +40,8 @@ class HostGenerator:
 
     _class_distribution = libiocage.lib.Distribution.DistributionGenerator
 
-    _devfs: 'libiocage.lib.DevfsRules.DevfsRules' = None
-    releases_dataset: libzfs.ZFSDataset = None
+    _devfs: 'libiocage.lib.DevfsRules.DevfsRules'
+    releases_dataset: libzfs.ZFSDataset
 
     def __init__(
         self,
--- a/libiocage/lib/Host.py
+++ b/libiocage/lib/Host.py
@@ -63,12 +63,17 @@ class HostGenerator:
             host=self,
             logger=self.logger
         )
-        self._defaults = defaults if defaults is not None else {}
+        self._defaults = defaults if defaults is not None \
+            else libiocage.lib.Resource.DefaultResource(
+                    dataset=self.datasets.root,
+                    logger=self.logger,
+                    zfs=self.zfs
+            )
 
     @property
     def defaults(self) -> 'libiocage.lib.Resource.DefaultResource':
         if self._defaults is None:
-            self._load_defaults()
+            self._defaults = self._load_defaults()
         return self._defaults
 
     @property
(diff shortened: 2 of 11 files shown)


libiocage/lib/Host.py
```diff
@@ -36,8 +36,8 @@ class HostGenerator:
 
     _class_distribution = libiocage.lib.Distribution.DistributionGenerator
 
-    _devfs: 'libiocage.lib.DevfsRules.DevfsRules' = None
-    releases_dataset: libzfs.ZFSDataset = None
+    _devfs: 'libiocage.lib.DevfsRules.DevfsRules'
+    releases_dataset: libzfs.ZFSDataset
```
igalic: What happens if we don't initialize a class variable?

libiocage/lib/ZFS.py
```diff
@@ -1,23 +1,16 @@
 import libzfs
 
-import libiocage.lib.Logger
-import libiocage.lib.helpers
-
 
 def get_zfs(
-    logger: 'libiocage.lib.Logger.Logger'=None,
```
igalic: it seems that in ZFS' case, `logger` is not used.

libiocage/lib/ZFS.py
```diff
@@ -49,4 +42,4 @@ def get_pool(self, name: str) -> libzfs.ZFSPool:
         for pool in self.pools:
             if pool.name == pool_name:
                 return pool
-        return None
+        raise
```
igalic: is this the best we can do???

libiocage/lib/helpers.py
```diff
@@ -44,16 +44,14 @@ def init_zfs(
 
     if (zfs is not None) and isinstance(zfs, libiocage.lib.ZFS.ZFS):
         object.__setattr__(self, 'zfs', zfs)
-        return zfs
+    else:
+        new_zfs = libiocage.lib.ZFS.get_zfs()
```
igalic: assign a new variable

libiocage/lib/helpers.py
```diff
@@ -44,16 +44,14 @@ def init_zfs(
...
+        object.__setattr__(self, 'zfs', new_zfs)
```
igalic: set the `zfs` attribute in both branches.

libiocage/lib/helpers.py
```diff
@@ -44,16 +44,14 @@ def init_zfs(
...
 
     try:
         return self.zfs
```
igalic: try to access the `zfs` attribute set above.

libiocage/lib/helpers.py
```diff
@@ -44,16 +44,14 @@ def init_zfs(
...
     except AttributeError:
-        pass
-
-    zfs = libiocage.lib.ZFS.get_zfs(logger=self.logger)
-    object.__setattr__(self, 'zfs', zfs)
-    return zfs
+        raise
```
igalic: raise if it fails

libiocage/lib/helpers.py
```diff
@@ -165,7 +163,7 @@ def validate_name(name: str) -> bool:
     return bool(_validate_name.fullmatch(name))
 
 
-def parse_none(data) -> str:
+def parse_none(data: typing.Any) -> None:
```
igalic: just telling the truth here;)

libiocage/lib/ZFS.py
gronke: We can return errors.ZFSPoolUnavailable

libiocage/lib/Host.py
```diff
@@ -63,7 +63,7 @@ def __init__(
             host=self,
             logger=self.logger
         )
-        self._defaults = defaults
+        self._defaults = defaults if defaults is not None else {}
```
gronke: Got your point! This should achieve the same without setting _defaults to a property other than None or DefaultResource.

```python
class HostGenerator:

    _class_distribution = libiocage.lib.Distribution.DistributionGenerator

    _devfs: 'libiocage.lib.DevfsRules.DevfsRules'
    _defaults: 'libiocage.lib.Resource.DefaultResource'
    releases_dataset: libzfs.ZFSDataset

    def __init__(
        self,
        root_dataset: libzfs.ZFSDataset=None,
        defaults: 'libiocage.lib.Resource.DefaultResource'=None,
        zfs: 'libiocage.lib.ZFS.ZFS'=None,
        logger: 'libiocage.lib.Logger.Logger'=None
    ) -> None:

        self.logger = libiocage.lib.helpers.init_logger(self, logger)
        self.zfs = libiocage.lib.helpers.init_zfs(self, zfs)

        self.datasets = libiocage.lib.Datasets.Datasets(
            root=root_dataset,
            logger=self.logger,
            zfs=self.zfs
        )
        self.distribution = self._class_distribution(
            host=self,
            logger=self.logger
        )

        if defaults is not None:
            self._defaults = defaults
```

libiocage/lib/Config/ConfigFile.py
```diff
@@ -29,11 +29,11 @@
 
 class ConfigFile(libiocage.lib.Config.Prototype.Prototype):
 
-    _file: str = None
+    _file: str
 
     def __init__(
         self,
-        file: str=None,
+        file: str,
```
gronke: `file` is explicitly allowed to be None. There are situations where the file is unknown. For example when a new jail gets created before name assignment.

gronke:
Your changes have been merged to master!

PR #99, opened 2017-09-16
install libzfs from pkg

gronke:
By installing py-libzfs from a FreeBSD/HardenedBSD package using pkg we no longer require to build Cython and to propagate parts of /usr/src.

@igalic we should give your solution to to partially download /usr/src upstream to py-libzfs 👍

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,13 +2,8 @@ ZPOOL=""
 SERVER=""
 SVN_REL_URL=https://svn.freebsd.org/base/releng/11.1
 
-/usr/src/cddl:
-	( cd /usr/src && svnlite checkout $(SVN_REL_URL)/cddl/ )
-/usr/src/sys/cddl:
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -2,13 +2,8 @@ ZPOOL=""
 SERVER=""
 SVN_REL_URL=https://svn.freebsd.org/base/releng/11.1
 
-/usr/src/cddl:
-	( cd /usr/src && svnlite checkout $(SVN_REL_URL)/cddl/ )
-/usr/src/sys/cddl:
-	( mkdir -p /usr/src/sys && cd /usr/src/sys && \
-		svnlite checkout $(SVN_REL_URL)/sys/cddl/ )
 install: /usr/src/cddl /usr/src/sys/cddl
```
igalic: we don't need this any more.

PR #101, opened 2017-09-16
Improve Typing

gronke:
Adds types to the following files:
- libiocage/lib/errors.py
- libiocage/lib/events.py

### DIFF
--- a/libiocage/lib/Types.py
+++ b/libiocage/lib/Types.py
@@ -23,10 +23,9 @@
 # POSSIBILITY OF SUCH DAMAGE.
 import typing
 import re
-from collections import UserString
 
 
-class AbsolutePath(UserString):
+class AbsolutePath(str):
     """Wrapper Type for ensuring a `str` matches a Unix Path"""
 
     unix_path = re.compile(r"/([^/\0]+/*)+")
@@ -37,16 +36,7 @@ class AbsolutePath(UserString):
     ) -> None:
         if self.unix_path.fullmatch(sequence) is None:
             raise TypeError("Invalid value for AbsolutePath")
-        self.data = sequence
-
-    def __str__(self) -> str:
-        return self.data
-
-    def __getitem__(self):
-        return str.__getitem__(self.data)
-
-    def __len__(self):
-        return str.__len__(self.data)
+        str.__init__(self, sequence)
 
 
 class UserInput:
--- a/libiocage/lib/Types.py
+++ b/libiocage/lib/Types.py
@@ -21,7 +21,7 @@
 # STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING
 # IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
 # POSSIBILITY OF SUCH DAMAGE.
-
+import typing
 import re
 from collections import UserString
 
@@ -38,3 +38,22 @@ class AbsolutePath(UserString):
         if self.unix_path.fullmatch(sequence) is None:
             raise TypeError("Invalid value for AbsolutePath")
         self.data = sequence
+
+    def __str__(self) -> str:
+        return self.data
+
+    def __getitem__(self):
+        return str.__getitem__(self.data)
+
+    def __len__(self):
+        return str.__len__(self.data)
+
+
+class UserInput:
+
+    def __init__(
+        self,
+        data: typing.Union[str, int, float, bool, None]
+    ) -> None:
+
+        ...
--- a/libiocage/lib/Types.py
+++ b/libiocage/lib/Types.py
@@ -1,0 +1,40 @@
+# Copyright (c) 2014-2017, iocage
+# All rights reserved.
+#
+# Redistribution and use in source and binary forms, with or without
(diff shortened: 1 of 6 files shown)


libiocage/lib/errors.py
```diff
@@ -22,17 +22,27 @@
 # IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
 # POSSIBILITY OF SUCH DAMAGE.
 
+# MyPy
+import libzfs  # noqa: F401
```
igalic: huh?

libiocage/lib/errors.py
```diff
@@ -22,17 +22,27 @@
...
 
 class IocageException(Exception):
 
-    def __init__(self, message, errors=None, logger=None, level="error",
-                 append_warning=False, warning=None):
+    def __init__(
+        self,
+        message: str,
+        logger: 'libiocage.lib.Logger.Logger'=None,
+        level: str="error",
```
igalic: this shouldn't be `str`, imo

libiocage/lib/errors.py
```diff
@@ -220,35 +267,59 @@ def __init__(self, distribution_name, *args, **kwargs):
 
 class UnmountFailed(IocageException):
 
-    def __init__(self, mountpoint, *args, **kwargs):
+    def __init__(
+        self,
+        mountpoint: str,
```
igalic: we should have a type for Unix paths

libiocage/lib/errors.py
```diff
@@ -220,35 +267,59 @@ def __init__(self, distribution_name, *args, **kwargs):
...
         msg = f"Failed to unmount {mountpoint}"
         super().__init__(msg, *args, **kwargs)
 
 
 class MountFailed(IocageException):
 
-    def __init__(self, mountpoint, *args, **kwargs):
+    def __init__(
+        self,
+        mountpoint: str,
```
igalic: a path type would be useful

libiocage/lib/events.py
```diff
@@ -70,15 +74,13 @@ def __init__(self, message=None, **kwargs):
...
-                         error="failed",
-                         skipped="skipped",
-                         done="done",
-                         pending="pending"):
+    def get_state_string(
+        self,
+        error: str="failed",
+        skipped: str="skipped",
+        done: str="done",
+        pending: str="pending"
```
igalic: once again, i don't like that these are `str`

libiocage/lib/events.py
```diff
@@ -92,7 +94,7 @@ def get_state_string(self,
         return pending
 
     @property
-    def type(self):
+    def type(self) -> str:
```
igalic: why._.

libiocage/lib/events.py
```diff
@@ -115,51 +117,51 @@ def pending(self, state):
...
         if new_state is False:
-            self._stopped_at = timer()
+            self._stopped_at = float(timer())
 
         self._pending = new_state
         IocageEvent.PENDING_COUNT += 1 if (state is True) else -1
 
     @property
-    def duration(self):
+    def duration(self) -> typing.Optional[float]:
```
igalic: why is this float? doesn't time come in better types?

libiocage/lib/Types.py
```diff
@@ -0,0 +1,40 @@
...
+
+    unix_path = re.compile(r"/([^/\0]+/*)+")
+
+    def __init__(
+            self,
+            sequence: str
+    ) -> None:
+        if self.unix_path.fullmatch(sequence) is None:
+            raise TypeError("Invalid value for AbsolutePath")
+        super.__init__(sequence)
```
igalic:
```
libiocage/lib/Types.py:40: error: No overload variant of "__init__" of "super" matches argument types [builtins.str]
```
ugh

libiocage/lib/errors.py
gronke: mypy complained about missing libzfs, so I've added it to the imports. Flake8 would now complain about the unused import unless we ignore F401.

libiocage/lib/Types.py
```diff
@@ -0,0 +1,40 @@
...
+# HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT,
+# STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING
+# IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+# POSSIBILITY OF SUCH DAMAGE.
+
+import re
+from collections import UserString
+
+
+class AbsolutePath(UserString):
```
igalic: I'm not sure `UserString` works here… i feel like the docs are lying to me… https://docs.python.org/3/library/collections.html#collections.UserString

libiocage/lib/events.py
gronke: this is the delta of two timestamps. what other than float do you suggest to use?

libiocage/lib/events.py
gronke: It's printed later. And it will be the reference for translations.

Issue #38, opened 2017-08-28
Bug: Do not **unintentionally** modify function arguments


igalic:
We have a handful of functions that take non-`None` arguments, and then modify them.
This leads to python directly modifying the passed argument, and is generally considered bad form.

One such example is `JailConfigAddresses#add()`

```python
    def add(self, nic, addresses=[], notify=True):

        if isinstance(addresses, str):
            addresses = [addresses]
```

maybe there are linter params that will warn us of such code, until then we might need human eyes.

gronke:
Agreed. Needs to be addressed! Should not be too complicated to fix, but a bit hard to spot. Is there a static code analyzer that does check for it?

igalic:
i don't know…
it doesn't seem to be in pylint, https://pylint.readthedocs.io/en/latest/technical_reference/extensions.html
and we already use flake8, which dosen't seem to catch those either: http://flake8.pycqa.org/en/latest/user/error-codes.html

igalic:
after installing `flake8-mutable`, i get the following:

```./iocage/lib/Network.py:35:33: M511 - mutable default arg of type List
./iocage/lib/Network.py:36:33: M511 - mutable default arg of type List
./iocage/lib/DevfsRules.py:33:47: M511 - mutable default arg of type List
./iocage/lib/Jail.py:186:63: M511 - mutable default arg of type Dict
./iocage/lib/Filter.py:46:36: M511 - mutable default arg of type Call
./iocage/lib/NetworkInterface.py:34:33: M511 - mutable default arg of type List
./iocage/lib/NetworkInterface.py:35:33: M511 - mutable default arg of type List
./iocage/lib/NetworkInterface.py:43:33: M511 - mutable default arg of type List
```

PR #107, opened 2017-09-17
flake8: check for mutable function parameters

igalic:
install flake8-mutable (and, just to be sure, flake8-builtins;) to check
for these things, as reported in #38.


igalic:
i think we can do better…

.travis/mypy-stubs/libzfs.pyi
```diff
@@ -54,21 +54,21 @@ class ZFS:
     datasets = ... # type: Any
     errno = ... # type: Any
     errstr = ... # type: Any
-    pools = ... # type: Any
+    pools: List[ZFSPool] = ...
```
igalic: This whole thing is amazing
thank you very much for this effort 💜

iocage/cli/set.py
```diff
@@ -66,15 +71,15 @@ def cli(ctx, props, jail):
 
     updated_jail_count = 0
 
-    for jail in ioc_jails:
+    for ioc_jail in ioc_jails:  # type: iocage.lib.Jail.JailGenerator
```
igalic: how the heck does that need a hint?? that feels like a bug in mypy

iocage/cli/set.py
```diff
@@ -107,6 +112,7 @@ def set_properties(
                 del target.config[key]
                 updated_properties.add(key)
             except:
+                raise
                 pass
```
igalic: this `pass` can probz be removed.

iocage/lib/Config/Jail/Defaults.py
```diff
@@ -26,11 +26,38 @@
 import iocage.lib.Config.Jail.BaseConfig
 
 
-class JailConfigDefaults(iocage.lib.Config.Jail.BaseConfig.BaseConfig):
+class DefaultsUserData(dict):
 
     user_properties: set = set()
 
-    data: dict = {
+    def __init__(self, defaults: dict={}) -> None:
```
igalic: doesn't this make `defaults` mutable?

iocage/cli/set.py
gronke: You can try to remove. Being explicit in other places should have fixed it anyway!

iocage/lib/Config/Jail/Defaults.py
igalic: not according to flake8-mutable.

iocage/lib/Config/Jail/Defaults.py
```diff
@@ -26,11 +26,38 @@
 import iocage.lib.Config.Jail.BaseConfig
 
 
-class JailConfigDefaults(iocage.lib.Config.Jail.BaseConfig.BaseConfig):
+class DefaultsUserData(dict):
```
igalic: can you briefly explain what happened.

iocage/lib/Datasets.py
```diff
@@ -151,38 +188,45 @@ def _set_zfs_property(self, dataset, name, value):
...
-        name = f"{root_name}/{name}"
+        dataset_name = f"{root_dataset_name}/{name}"
         try:
-            dataset = self.zfs.get_dataset(name)
+            dataset = self.zfs.get_dataset(dataset_name)
         except:
-            pool.create(name, {})
-            dataset = self.zfs.get_dataset(name)
+            target_pool.create(dataset_name, {})
+            dataset = self.zfs.get_dataset(dataset_name)
```
igalic: thanks ;)

iocage/lib/Jail.py
```diff
@@ -218,6 +232,22 @@ def __init__(
...
+        JailResource.__init__(
+            self,
+            jail=self,
+            host=self.host,
+            logger=self.logger,
+            zfs=self.zfs,
+            **resource_args
+        )
+
+        if not new and (("id" not in data) or (data["id"] is None)):
```
igalic: this looks like a candidate for encapsulation ;)

iocage/lib/LaunchableResource.py
```diff
@@ -69,17 +72,6 @@ def dataset_name(self, value: str) -> None:
...
-        if self._dataset is None:
-            self._dataset = self.zfs.get_dataset(self.dataset_name)
-
-        return self._dataset
-
-    @dataset.setter
-    def dataset(self, value: libzfs.ZFSDataset):
-        self._set_dataset(value)
-
-    @property
```
igalic: what happened here?

iocage/lib/Config/Jail/Defaults.py
gronke: JailConfigDefaults should return all default properties as `data`, but only the user overrides as `user_data`, so that only manual overrides are saved to `defaults.json`. It was not sufficient to provide a reduced copy of self.data as user_data because some properties were references. DefaultsUserData goes one level deeper and intercepts setting individual attributes, so that we no longer need to pass the reduced copy of data.

iocage/cli/set.py
igalic: you're right
works without it, too.

iocage/lib/LaunchableResource.py
gronke: It's implemented by the parent class. This overrides are no longer needed.

iocage/lib/Jail.py
gronke: It's below 80 characters!!!1 😼

igalic: yea, but, the logic is too much for my tired brain 😩

iocage/lib/LaunchableResource.py
igalic: aye

PR #108, opened 2017-09-17
remove fastentrypoints & simplify code without it

igalic:
we can do without it, so lets.


skarekrow:
Why? It speeds up execution.

igalic:
because it's an optimization we don't need or use
(according to the helpful people in #python. and as such, pointless)

skarekrow:
https://github.com/iocage/iocage/issues/72 this was the original issue, and it was certainly useful then. I can't imagine what changed the situation.

igalic:
i love how this whole PR is a perfect example for [Chsterton's Fence](https://en.wikipedia.org/wiki/Wikipedia:Chesterton%27s_fence)

ninjaaron:
lol

Issue #111, opened 2017-09-19
analyse libiocage code for security sensitive issues


igalic:
being a security-sensitive project, we should consider _automatically_ auditing our code for potential security issues.
being a security conscious project, i suggest we _"simply"_ add [bandit](https://pypi.python.org/pypi/bandit) to our travis tests.

igalic:
i've tested the current master branch with

```
bandit --exclude iocage/tests/ -r .
```

and it looks quite _okay_
two of the high-severity issues are false-positives, since we're on Python 3.x.
The rest should be easy to clean up.

gronke:
Is there a way to integrate this kind of check in our build system?

igalic:
same as with mypy and flake8:

```diff
diff --git a/.travis.yml b/.travis.yml
index a45dc69..61e1dc6 100644
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,7 @@ sudo: false
 env:
   - MYPYPATH="$TRAVIS_BUILD_DIR/.travis/mypy-stubs"
 install:
-  - pip3 install flake8-mutable flake8-builtins flake8-mypy
+  - pip3 install flake8-mutable flake8-builtins flake8-mypy bandit
 script:
   - flake8 --exclude=__init__.py --ignore=E203,W391
+  - bandit --exclude iocage/tests/ -r .
diff --git a/requirements-dev.txt b/requirements-dev.txt
index 1917390..013ad27 100644
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -2,5 +2,8 @@ pytest
 pytest-cov
 pytest-pep8
 pytest-mock
-mypy
+flake8-mutable
+flake8-builtins
+flake8-mypy
+bandit
 git+https://github.com/freenas/py-libzfs.git#egg=libzfs
```

gronke:
Bandit was embedded in the build system with PR #112

PR #112, opened 2017-09-19
install bandit and start "security" scannig our code

igalic:
as discussed in #111, we would like to analyze our code for potential
security issues. a first step here is to install a code scanner and
start using it!


skarekrow:
I don’t see the value in doing this. Most of the commit is adding a comment to skip the scanning in the first place. I can only imagine how many of those comments we’re going to have. Why are we using bandit?

araujobsd:
I will give my 0.50 cents! Instead of do a make up the issue, would be better let it exposed till somebody can real fix it, in case it is a real issue.

igalic:
the value is to be aware of what we're doing. `# nosec: Yes, we want a shell-call here` is vastly different from `# nosec`, for one.

For the other, we don't need to keep it:

Right now, libiocage is exeperimental, and i hope no one is using it for their production systems, we can still — openly — use an automated system — such as bandit, or maybe something else — to tease out the issues, or at least, become aware of them.

For instance, https://github.com/iocage/libiocage/issues/113 is a consequence of this PR.

gronke:
Bandit is not relevant for runtime execution, so it won't affect the enduser. It's better for us to have strict analysis tools in place to give people working on pull-requests a better chance to fit the project requirements. The tighter we define the code quality requirements on our build system, the easier it becomes to merge code contributions.

> the value is to be aware of what we're doing. # nosec: Yes, we want a shell-call here is vastly different from # nosec, for one.

I absolutely want to be explicit with dangerous things like forks. 💛

araujobsd:
Perhaps an additional section at README with suggestion of tools to be used before submit a PR makes more sense than enforce tools usage such like: mypy and/or bandit.

Or otherwise in a near future, if a submitter sends a patch and he is not aware of these tools he would feel frustrated and probably will go away.

I strongly suggest to create an additional section at README and expose these tools, before pick up a tool and enforce the usage. At least everybody will be aware of the tools the project is choosing.

As an additional example: travis will be always broke and soon or late, devs will start to ignore it, what is dangerous.

gronke:
> I strongly suggest to create an additional section at README and expose these tools, before pick up a tool and enforce the usage. At least everybody will be aware of the tools the project is choosing.

Yes. And in addition we should break the various checks into multiple Travis tasks. The total execution time will be longer, but on a PR it would show exactly which metric does not fit the requirements. We can always point a link to a description in our README. Good call, @araujobsd

skarekrow:
+1

igalic:
travis has been broken for weeks on every pull request that @gronke and myself have worked on, ***because*** we've been working on pull requests to improve the code-quality with the help of automated testing tools

we haven't merged anything that was broken, so master has been green (we need to reflect that as a badge in our README ;)

i absolutely agree with you on helping contributors submit PRs that will be travis-green from the get-go. To do that, we should add a `make check` (and reuse that in `.travis.yml`)

araujobsd:
My suggestion would be, first start with the additional section listing all tools the project intend to use as best practices. After a while and after it gets mature among devs, it will be natural the adoption of it and then enforce it.

Overall all these tools have something good, just need apply in homeopathic doses.

igalic:
switching gears, and focusing on the *results* of this PR so far:

most of our issues are currently [B110: try_except_pass](https://docs.openstack.org/bandit/latest/plugins/try_except_pass.html)

that is, code of this form:

```python
try:
  do_some_stuff()
except Exception:
  pass
```

we should consider refactoring this pattern to at least catch the _"correct"_ exception.

araujobsd:
I agree with the exception, it should catch the right exception and do something not only pass all exceptions!  Although sometimes it is inevitable and it is a necessary evil.

igalic:
Another class of issues currently marked with medium severity is _possibly_ a bug in bandit: [[B310:blacklist] Audit url open for permitted schemes.](https://docs.openstack.org/bandit/latest/api/bandit.blacklists.html#b310-urllib-urlopen) fails to notice that our `mirror_url` `@setter` already does the "audit" for schemes as bandit recommends!

I cannot seem to find anything related in [bandit's issue tracker (laundpad!)](https://bugs.launchpad.net/bandit), so if someone knows any of its developers, it would be great to point them here.

iocage/cli/__init__.py
```diff
@@ -25,7 +25,7 @@
 import os
 import re
 import signal
-import subprocess as su
+import subprocess  # nosec: B804
```
igalic: this should be B404, shouldn't it?
i disabled that entirely in the makefile, and tried disabling it in setup.cfg

iocage/lib/Config/Jail/Properties/Resolver.py
```diff
@@ -21,6 +21,7 @@
 # STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING
 # IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
 # POSSIBILITY OF SUCH DAMAGE.
+import typing
```
igalic: this should be a separate commit

setup.cfg
```diff
@@ -1,6 +1,9 @@
 [metadata]
 description-file = README.md
 
+[bandit]
+skip = B404
```
igalic: does this actually work

gronke: It did complain about B404, so I don't think so.

iocage/cli/__init__.py
gronke: aye! will update

gronke: Just figured out that despite of a defined exception, `# nosec: <ANYTHING>` does skip all bandit warnings for this line. We should be careful with that and fix upstream

gronke:
> I don’t see the value in doing this. Most of the commit is adding a comment to skip the scanning in the first place.

@skarekrow the errors reported by bandit have been valuable to know. It found for example swallowed exceptions where we didn't precisely defined the expected errors. It's a nice addition to MyPy and should be required in the build process.

PR #117, opened 2017-09-22
remove dependency on libgit2

gronke:
We did never use this library, so it can be removed.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,7 +2,7 @@ ZPOOL=""
 SERVER=""
 
 install:
-	pkg install -q -y libgit2 libucl cython3 rsync python36 py36-libzfs
+	pkg install -q -y libucl cython3 rsync python36 py36-libzfs
 	python3.6 -m ensurepip
 	pip3.6 install -Ur requirements.txt
 	pip3.6 install -e .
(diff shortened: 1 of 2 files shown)


igalic:
from what i gather, iocage did use it, but libiocage did not.

skarekrow:
Correct :)

gronke:
If the dude who reported the Hardened BSD issue to me reads this: Get yourself a GitHub account - your feedback is valuable!

gronke:
What about cffi? We should carefully check all dependencies and see what can be removed or simplified!

@igalic do you know a good way to measure code complexity with mypy? Think that metric will help us finding weak spots.

igalic:
i honestly think i'm a very good measure for complexity 😝 
if i complain about too deep `if`s then plz fix them!

there's https://github.com/Miserlou/JonesComplexity — which we already practice thanks to our adherence to <80 char per line, while also trying to have the longest class hierarchies ever seen. (this module doesn't install right now anyway, because `ModuleNotFoundError: No module named 'pypandoc'`

and then there's https://pypi.python.org/pypi/mccabe — "the classic"
which, when ran over our current source base, finds, nothing.

PR #118, opened 2017-09-22
fix pool activation

gronke:
The initialization of iocage.lib.Datasets.Datasets required an activated pool, but was involved in the activation process. With this change Datasets can be initialized without an active pool, but would fail on dataset access instead.


iocage/lib/Datasets.py
```diff
@@ -52,32 +52,45 @@ def __init__(
...
+        pool = self._active_pool_or_none
+        if pool is None:
+            raise iocage.lib.errors.IocageNotActivated(logger=self.logger)
+        return pool
+
+    @property
+    def root(self):
+        try:
+            return self._root
+        except:
```
igalic: nope

PR #125, opened 2017-09-23
Fix various issues in fstab

gronke:
- prepend basejail nullfs mounts
- keep order of custom fstab entries
- bulk unmount nullfs mounts

### DIFF
--- a/iocage/lib/Jail.py
+++ b/iocage/lib/Jail.py
@@ -991,29 +991,33 @@ class JailGenerator(JailResource):
 
     def _teardown_mounts(self) -> None:
 
-        mountpoints = list(map(
-            self._get_absolute_path_from_jail_asset,
-            [
-                "/dev/fd",
-                "/dev",
-                "/proc",
-                "/root/compat/linux/proc"
-            ]
+        mountpoints = list(filter(
+            os.path.isdir,
+            map(
+                self._get_absolute_path_from_jail_asset,
+                [
+                    "/usr/bin",
+                    "/dev/fd",
+                    "/dev",
+                    "/proc",
+                    "/root/compat/linux/proc"
+                ]
+            )
         ))
 
-        mountpoints += list(map(
-            lambda x: x["destination"],
-            list(self.fstab)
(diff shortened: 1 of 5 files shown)


iocage/lib/Jail.py
```diff
@@ -991,29 +991,32 @@ def update_jail_state(self) -> None:
...
-        for mountpoint in mountpoints:
-            if os.path.isdir(str(mountpoint)):
-                iocage.lib.helpers.umount(
-                    mountpoint,
-                    force=True,
-                    logger=self.logger,
-                    ignore_error=True  # maybe it was not mounted
-                )
+        iocage.lib.helpers.umount(
+            ["-A", "-F", self.fstab.path, "-t", "nullfs"],
```
igalic: can you please explain the options?

iocage/lib/LaunchableResource.py
```diff
@@ -46,14 +46,23 @@ def create_resource(self) -> None:
...
+            )
+
     @property
-    def root_path(self):
+    def root_path(self) -> str:
         return self.root_dataset.mountpoint
 
     @property
     def root_dataset(self) -> libzfs.ZFSDataset:
         # ToDo: Memoize root_dataset
```
igalic: TODO: open a bug for all TODOs we have

gronke: > TODO: open a bug for all TODOs we have

for *each* ToDo 😎

iocage/lib/Jail.py
gronke: `-A` unmounts all entries from fstab
`-F` sets the provided path as fstab
`-t nullfs` ensures we only unmount nullfs mounts

igalic: what about potential entries of other types?

gronke: Finally you pointed me to an issue. It's important not to use `-A`, but `-a`. The latter removes the entries from fstab - `-A` removes all from the system :trollface:

PR #136, opened 2017-09-24
fix ZFS Share feature on FreeBSD

gronke:
Currently there are various issues mounting ZFS shares. This Pull-Request aims to fix and manually test them.


gronke:
I could not get ZFS shares to work on HardenedBSD yet - needs investigation.

PR #137, opened 2017-09-24
Feature: Custom Releases

gronke:
- document steps to create custom release from /usr/src
- expose fetch flag to update a base release only

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,6 +49,25 @@ jail.create("11.1-RELEASE")
 
 Libiocage comes bundles with a CLI tool called `ioc`. It is inspired by the command line interface of [iocage](https://github.com/iocage/iocage) but meant to be developed along with the library and to spike on new features.
 
+### Custom Release (e.g. running -CURRENT)
+
+#### Initially create release dataset
+
+```sh
+zfs create zroot/iocage/releases/custom/root
+cd /usr/src
+#install your source tree
+make installworld DESTDIR=/iocage/releases/custom/root
+make distribution DESTDIR=/iocage/releases/custom/root
+ioc fetch -r custom -b
+```
+
+#### Update the installation after recompile
+```sh
+make installworld DESTDIR=/iocage/releases/custom/root
+ioc fetch -r custom -b
+```
+
 ## Development
 
 ### Install Development Dependencies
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -49,6 +49,25 @@ jail.create("11.1-RELEASE")
 
 Libiocage comes bundles with a CLI tool called `ioc`. It is inspired by the command line interface of [iocage](https://github.com/iocage/iocage) but meant to be developed along with the library and to spike on new features.
 
+### Custom Release (e.g. running -CURRENT)
```
igalic: so from what i gather, this is a way of customizing a release, a.k.a: creating something that we previously might have called a template!

README.md
```diff
@@ -49,6 +49,25 @@ jail.create("11.1-RELEASE")
...
+
+#### Initially create release dataset
+
+```sh
+zfs create zroot/iocage/releases/custom/root
```
igalic: how about storing a configuration in `zroot/iocage/releases/custom/config.json` ?
that details what this release is cut from (in case we need to `freebsd-update` it) and which directories need to be mounted into a jail that's based on it.

gronke: good call!

README.md
gronke: I was unsure whether to make it feature of iocage or document how to do manually. This seemed to be the simplest to allow template-ish acrobatics.

igalic: if we add it as feature, it's the officially encouraged method, if we document it, people will keep asking us how to it.
if we make the feature general enough, people can build on top of it, if we don't offer anything other than documentation, people will keep reinventing wheels that we'll have to support

README.md
foo2342: This feature in not a replacement for templates, but rather a way to run unsupported releases, e.g. -STABLE or -CURRENT. For these installations freebsd-update is not possible.
Updates are simply done by writing the binaries and libraries over the release, not touching the config files, which are written by make distribution.
This approach is used by ezjail as well.

gronke:
@igalic I agree that a solid implementation is better than just documenting the steps. We will re-visit this topic one more time when it comes to templates. I'd like to introduce the cli/fetch flag to exclusively update a basejail dataset from it's release because it gives the power to the user until we come up with a baked in solution.

PR #140, opened 2017-09-25
Feature: USB Support

gronke:
- allows to expose USB devices via `allow_usb` configuration
- Define individual device(s)

### DIFF
--- a/libioc/Config/Jail/BaseConfig.py
+++ b/libioc/Config/Jail/BaseConfig.py
@@ -536,6 +536,19 @@ class BaseConfig(dict):
         except KeyError:
             return "local"
 
+    def _get_usb_device(self) -> typing.List[str]:
+        devices = self.data["usb_device"].split()  # type: typing.List[str]
+        return devices
+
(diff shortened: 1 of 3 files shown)


iocage/lib/Jail.py
```diff
@@ -673,9 +673,15 @@ def devfs_ruleset(self) -> iocage.lib.DevfsRules.DevfsRuleset:
         devfs_ruleset = iocage.lib.DevfsRules.DevfsRuleset()
         devfs_ruleset.clone(configured_devfs_ruleset)
 
-        if self._dhcp_enabled:
+        if self._dhcp_enabled is True:
             devfs_ruleset.append("add path 'bpf*' unhide")
 
+        if self.config["allow_usb"] is True:
+            devfs_ruleset.append("add path 'usb/*' unhide")
```
gronke: Need to investigate the security impact by unhiding `/dev/usb/*`

igalic: why is this even necessary after the lines below?

Issue #155, opened 2017-10-01
StandaloneJailStorage has no attribute logger


igalic:
while cleaning up #131, we noticed that the class `StandaloneJailStorage` uses `logger` without having it declared anywhere, or inherited from anything.

****

Full set of errors:

```
iocage/lib/StandaloneJailStorage.py:27:4: error: Function is missing a type annotation for one or more arguments
iocage/lib/StandaloneJailStorage.py:28:8: error: "StandaloneJailStorage" has no attribute "logger"
iocage/lib/StandaloneJailStorage.py:30:17: error: "StandaloneJailStorage" has no attribute "jail"
iocage/lib/StandaloneJailStorage.py:33:4: error: Function is missing a type annotation for one or more arguments
iocage/lib/StandaloneJailStorage.py:35:8: error: "StandaloneJailStorage" has no attribute "logger"
iocage/lib/StandaloneJailStorage.py:36:8: error: "StandaloneJailStorage" has no attribute "clone_release"
```

PR #158, opened 2017-10-03
types for Storage classes

gronke:
#155 points out missing type interfaces in Storage classes.


iocage/lib/StandaloneJailStorage.py
```diff
@@ -24,13 +24,21 @@
 
 
 class StandaloneJailStorage:
-    def apply(self, release):
+    
+    def apply(
+        self,
+        release: 'iocage.lib.Release.ReleaseGenerator'
```
igalic: importing `iocage.lib.Release` into this file would be too much?

iocage/lib/StandaloneJailStorage.py
```diff
@@ -24,13 +24,21 @@
...
+    ) -> None:
+
         self.logger.warn(
```
igalic: `logger` is still not part of this class

Issue #159, opened 2017-10-04
use yapf for auto-indentation


igalic:
rather than having a lot of text in CONTRIBUTING.md for exactly how to indent our code, we should use [yapf](https://github.com/google/yapf), now that it finally supports python 3.6!

skarekrow:
I’ve been using it in iocage. It’s great!

igalic:
@skarekrow what config are you using?
the one i use in #160 produced a giant diff that looks different form our style?
would you mind helping with that PR?

gronke:
Seems to be a good candidate for a make command like `make beautiful` or `make the code look better`.

@skarekrow what were your learnings from implementing it for iocage?

skarekrow:
I agree that it is a good candidate for that. As for learnings, not many. I let my editor do it with [ALE](https://github.com/w0rp/ale), instead of manually running it with any rules.

PR #160, opened 2017-10-04
chore(indent): use yapf for (auto) indentation

igalic:
This patch addresses #159.
However, we'll need to wait unil https://github.com/google/yapf/issues/452
is fixed before we can get started for real.


gronke:
Rebased and conflicts resolved.

@igalic have you seen [the reason](https://github.com/google/yapf/issues/452#issuecomment-368558051) why google/yapf#452 is closed? Can we get for real? And could you describe in the Readme how to use this?

Issue #14, opened 2017-08-22
Feature: Templates


igalic:
currently, libiocage doesn't implement Templates!

mostly, because, conceptually, it's really hard to grasp what they are, and how to implement them in a way that's intuitive, performant, and convenient.

@gronke and me discussed mapping it to the existing jail types, and it seems the most sensible thing would be to make them akin to NullFS BaseJails.

skarekrow:
They're not a hard concept really. You set up a jail how you would like it, and mark it as a template. Then you create from that template instead of a RELEASE.

That I believe is already intuitive, performant and convenient. 

I believe that simply setting a jail to `readonly` and changing it's type should be enough to distinguish these from jails. Alternatively since it's slightly redundant to clone jails, we can just drop templates and have users clone.

igalic:
yeah, i was thinking of implementing the _concept_ of templates in [puppet-jail](https://github.com/xaque208/puppet-jail) even if there was no hard distinguished feature.

HOWEVER, it's much much easier — for management purposes to distinguish templates, if the `type` is actually `template`.

skarekrow:
I agree, so my suggestion is:
```
type = template
readonly = on
```
That should be all that's needed to distinguish a template. We have legacy users using `POOL/iocage/templates` but they could easily be moved to this system when they choose by setting `templates=no` and back to `yes`.

igalic:
you wouldn't move those jails on iocage upgrade?

skarekrow:
Why would they be moved on upgrade? That's for jails, not templates. Templates should be immutable. So that means no starting, or modifying them in anyway.

Upgrade shouldn't change the structure of the filesystem outside the jail, that's way out of scope.

igalic:
what does that mean if i want to upgrade all the jails derived from a certain template?
If they were NullFS based, it could mean:

* destroy template
* rebuild template from new RELEASE
* reinstall needed packages
* …
* set `type = template` & `readonly = on`

and now, we can reboot our jails, and they'll be up-to-date.

Does this make sense?

skarekrow:
No that doesn't make sense.

You don't need to destroy the template, you simply set `template` to `no`. Then it's a jail you can change, and when you're happy set it back to `yes`. Instead of requiring the user to set both of those properties, that one will handle that internally. 

Destroying the template means destroying the jails, as they are children of that template.

[...]

PR #168, opened 2017-10-07
Feature: Templates

gronke:
implements #14 

- [x] allow to flag jails as template
- [x] `ioc list -t` lists templates
- [x] a template may not be started
- [x] create new jails from template
- [ ] update existing jails from a template *

* requires usage of `rsync --compare-dest` as commented [here](https://github.com/iocage/libiocage/issues/14#issuecomment-334738301)

### Usage
```
ioc create -n mytemplate
ioc start mytemplate
ioc console mytemplate
mytemplate$ # ... install things
mytemplate$ exit
ioc set template=yes mytemplate
ioc list -t
# verify the template is listed
ioc create -b -t mytemplate -n somejail
# done
```

### DIFF
--- a/iocage/cli/create.py
+++ b/iocage/cli/create.py
@@ -87,7 +87,7 @@ def cli(ctx, release, template, count, props, pkglist, basejail, basejail_type,
 
     jail_data = {}
 
-    if release is None:
+    if (release is None) and (template is None):
         logger.spam(
             "No release selected (-r, --release)."
             f" Selecting host release '{host.release_version}' as default."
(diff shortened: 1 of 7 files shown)


gronke:
@rwestlund this is a first step toward the approaches we discussed. Manual testing showed that rsync is a working solution for even updating jails from a template, but I'd like to dig into that after the planned alpha release for manual testing. IMO we can only withstand the introduced complexity with proper unit tests.

iocage/cli/create.py
```diff
@@ -97,30 +97,41 @@ def cli(ctx, release, template, count, props, pkglist, basejail, basejail_type,
...
         )
-        if no_fetch:
-            logger.error(msg)
-            exit(1)
-        else:
-            logger.spam(msg)
-            logger.log("Automatically fetching release '{release.name}'")
-            release.fetch()
+    else:
+        logger.error("No release or jail selected")
```
igalic: template, not jail

iocage/lib/Config/Jail/BaseConfig.py
```diff
@@ -68,13 +68,14 @@ class BaseConfig(dict):
...
     )
-    data: dict = {}
+    data: dict
 
     def __init__(
         self,
         logger: 'iocage.lib.Logger.Logger'=None
     ) -> None:
 
+        self.data = {}
```
igalic: what's the difference here?

iocage/lib/Jail.py
```diff
@@ -351,7 +351,7 @@ def _run_hook(self, hook_name: str):
             f"Running {hook_name} hook for {self.humanreadable_name}"
         )
 
-        lex = shlex.shlex(value)
+        lex = shlex.shlex(value)  # noqa: T484
```
igalic: oh?

iocage/lib/Jail.py
```diff
@@ -956,6 +982,28 @@ def _configure_route(self, gateway, ipv6=False) -> None:
...
         self.exec(command)
 
+    def require_jail_is_template(self, **kwargs) -> None:
+        """
+        Raise JailIsTemplate exception if the jail is a template
+        """
+        if self.config['template'] is False:
+            raise iocage.lib.errors.JailNotTemplate(
+                jail=self,
+                logger=self.logger,
```
igalic: wait, why do our exceptions need logger??

iocage/lib/Config/Jail/BaseConfig.py
gronke: Previously it was a class variable, now it's an instance variable. Makes a difference when working with multiple jails (like we do when using templates)

iocage/lib/Jail.py
gronke: This started spitting in all our branches. We can remove later, but it was blocking my work.

My plan is to revisit all `# noqa` occasions before we call out an alpha release.

iocage/lib/Jail.py
gronke: Their job is to produce log entries. They print with a high loglevel, so that we do not need to care about those error messages in the CLI. We can just exit(1) on IocageException.

iocage/lib/Jail.py
igalic: 😎

PR #171, opened 2017-10-08
Feature: cli/restart

gronke:
Adds a `restart` command to the CLI.


gronke:
Other iocage versions support soft-restarts by executing the exec_stop and exec_start scripts -
 needs to be implemented here as well to complete the feature.

PR #182, opened 2017-10-23
Enhancement: Call jls command only once for lists of jails

gronke:
- adds JailState data structure
- the Jails list iterator invokes update of jails state with jls only once and passes fitting results to jails as it iterates them

A quick comparison with 10 jails on SSD:
- Before: 2.51s real
- After: 0.6s real

### DIFF
--- a/iocage/lib/Jail.py
+++ b/iocage/lib/Jail.py
@@ -22,6 +22,7 @@
 # IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
 # POSSIBILITY OF SUCH DAMAGE.
 import typing
+import json
 import os
 import subprocess  # nosec: B404
 import uuid
@@ -1055,7 +1056,6 @@ class JailGenerator(JailResource):
         Invoke update of the jail state from jls output
         """
         try:
-            import json
             stdout = subprocess.check_output([
                 "/usr/sbin/jls",
                 "-j",
--- a/iocage/lib/Jail.py
+++ b/iocage/lib/Jail.py
@@ -462,7 +462,7 @@ class JailGenerator(JailResource):
             self.logger.verbose(
                 f"Dataset {current_dataset_name} renamed to {new_dataset_name}"
             )
-        except:
+        except BaseException:
             self.config["id"] = current_id
             raise
 
@@ -477,7 +477,7 @@ class JailGenerator(JailResource):
 
         try:
             self._run_hook("prestop")
-        except:
+        except Exception:
             self.logger.warn("pre-stop script failed")
 
         yield jailDestroyEvent.begin()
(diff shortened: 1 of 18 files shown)


iocage/cli/start.py
```diff
@@ -98,6 +98,7 @@ def start_jails(jails, logger, print_function):
             jail.require_jail_not_template()
             print_function(jail.start())
         except iocage.lib.errors.IocageException:
+            raise
             failed_jails.append(jail)
```
igalic: how's this supposed to work?

iocage/cli/stop.py
```diff
@@ -86,7 +86,7 @@ def stop_jails(
     for jail in jails:
         try:
             print_function(jail.stop(force=force))
-        except iocage.lib.errors.IocageException:
+        except iocage.lib.errors.IocageException as e:
```
igalic: you're not using `e`

iocage/lib/Jail.py
```diff
@@ -450,7 +462,7 @@ def rename(self, new_name: str) -> None:
             self.logger.verbose(
                 f"Dataset {current_dataset_name} renamed to {new_dataset_name}"
             )
-        except:
+        except BaseException:
```
igalic: what's `BaseException`?

iocage/lib/JailState.py
```diff
@@ -0,0 +1,126 @@
...
+        self,
+        states: typing.Optional[JailStatesDict]=None
+    ) -> None:
+
+        if states is None:
+            dict.__init__(self, {})
+        else:
+            dict.__init__(self, states)
+
+    def query(self):
```
igalic: is this a standard `dict` function here?

iocage/lib/Jails.py
```diff
@@ -76,8 +77,32 @@ def _create_resource_instance(
...
+            self.logger.spam(
+                f"Injecting pre-loaded state to '{jail.humanreadable_name}'"
+            )
+            jail.jail_state = self.states[jail.identifier]
+
         return jail
 
+    def __iter__(
+        self
+    ) -> typing.Generator['iocage.lib.Resource.Resource', None, None]:
```
igalic: is this necessary to be a `'ForwardReference'`?

iocage/lib/Jail.py
gronke: All Python Exceptions inherit from BaseException.

iocage/lib/JailState.py
gronke: no, but update (my previous naming choice) was. This is called to query the state from jls.

iocage/lib/Jails.py
gronke: No, I don't see the requirement, but passed it as string for convenience with other situations where it makes a difference. (Idea for the future: We could aim to refactor all circular dependencies to a linear dependency tree.)

PR #200, opened 2017-11-11
fix release updates on hosts with custom kernel

gronke:
When a custom kernel was compiled on a host that attempts to fetch and install release updates in a jail, it mistakenly assumes kernel debug modules. Since we are running jails only, there is no need to patch the kernel separately.

@igalic thanks for pairing on this!

### DIFF
--- a/iocage/lib/Release.py
+++ b/iocage/lib/Release.py
@@ -623,6 +623,7 @@ class ReleaseGenerator(ReleaseResource):
                         yield event
                     else:
                         changed = event
+            jail.stop()
             yield runReleaseUpdateEvent.end()
         except Exception as e:
             # kill the helper jail and roll back if anything went wrong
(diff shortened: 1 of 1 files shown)


iocage/lib/Release.py
```diff
@@ -531,10 +532,11 @@ def fetch_updates(
                 if self.host.distribution.name == "FreeBSD":
                     with open(local_path, "r+") as f:
                         content = f.read()
+                        pattern = re.compile("^Components .+$", re.MULTILINE)
```
igalic: my non-existent horse for a world where we can do string replacement without regex…

iocage/lib/Release.py
```diff
@@ -589,6 +591,7 @@ def update(
 
         jail = iocage.lib.Jail.JailGenerator(
             {
+                "name": self.name.replace(".", "-"),
```
igalic: this commit needs a bit more elaboration…
what's the benefit of giving it a name, why are '.' bad?

PR #223, opened 2017-11-26
Feature: Secure VNET

gronke:
When multiple jails share a bridge device, a malicious jail may spoof its IP or MAC address. In a common network scenario, where the host bridges multiple IP addresses from the same uplink interface on a shared bridge device, the VNET jails may pretend to own any of the available addresses.

This feature uses a second epair device pair and an additional bridge to enforce a jails MAC and IP address and to apply additional filters with pf.

### Current VNET

1. HOST_UPLINK vtnet0
2. PUBLIC_BRIDGE (bridge1)
3. EPAIR_A (Host, vnet0:23)
4. EPAIR_B (Jail, vnet0)

### Secured VNET

1. HOST_UPLINK vtnet0
2. PUBLIC_BRIDGE (bridge1)
3. EPAIR_C (Host, **vnet0:23:a**)
5. EPAIR_D (Host, **vnet0:23:b**)
5. JAIL_BRIDGE (Host, **vnet0:23:net**)
6. EPAIR_A (Host, vnet0:23)
7. EPAIR_B (Jail, vnet0)

---

Introduces configuration syntax for pointopoint gateway routes as required to bridge in some network situations. For example:

```sh
ADDITIONAL_IP=10.23.42.2
GATEWAY=`netstat -4 -rn | grep ^default | awk '{ print $2 }'`
MAC_HETZNER_IP="00:53:00:10:C4:63"
MAC_RANDOM="00:53:0B:AD:C0:DE"

ioc set vnet0=on \
  interfaces="vnet0::bridge0" \
  vnet0_mac="$MAC_HETZNER_IP,$MAC_RANDOM" \
  ip4_add="vnet0|$ADDITIONAL_IP/32" \
  defaultrouter="$GATEWAY@vnet0" \
  my-jail
```

---

Currently there is no firewall support included. This means the jails are still able to spoof source addresses, but we have an anchor point to hook in layer-2 filtering.

### DIFF
--- a/iocage/lib/Config/Jail/Properties/Interfaces.py
+++ b/iocage/lib/Config/Jail/Properties/Interfaces.py
@@ -96,7 +96,7 @@ class InterfaceProp(dict):
         """
         try:
             bridge = iocage.lib.helpers.parse_none(bridge_if)
-        except:
+        except TypeError:
             bridge = iocage.lib.BridgeInterface.BridgeInterface(bridge_if)
 
         dict.__setitem__(self, jail_if, bridge)
@@ -118,9 +118,8 @@ class InterfaceProp(dict):
     def __notify(self) -> None:
         self.config.update_special_property(self.property_name)
 
-    def __empty_prop(self, key: str) -> str:
+    def __empty_prop(self, key: str) -> None:
         dict.__setitem__(self, key, None)
-        return None
 
     def to_string(self, value: dict) -> str:
         out = []
--- a/iocage/lib/Config/Jail/Properties/Interfaces.py
+++ b/iocage/lib/Config/Jail/Properties/Interfaces.py
@@ -26,36 +26,6 @@ import typing
 import iocage.lib.helpers
 
 
-class BridgeSet(set):
-
-    config: 'iocage.lib.Config.Jail.JailConfig.JailConfig'
-
-    def __init__(
-        self,
-        config: typing.Optional[
-            'iocage.lib.Config.Jail.JailConfig.JailConfig'
-        ]=None
-    ) -> None:
-
-        if config is not None:
-            self.config = config
-
-        set.__init__(self)
-
-    def add(self, value: str, notify: bool=True) -> None:
-        set.add(self, value)
-        if notify:
-            self.__notify()
-
-    def remove(self, value: str, notify: bool=True) -> None:
-        set.remove(self, value)
-        if notify:
-            self.__notify()
-
-    def __notify(self) -> None:
-        self.config.update_special_property("interfaces")
-
-
 class InterfaceProp(dict):
 
     config: 'iocage.lib.Config.Jail.JailConfig.JailConfig'
@@ -76,7 +46,7 @@ class InterfaceProp(dict):
 
     def set(
         self,
-        data: typing.Union[str, typing.Dict[str, BridgeSet]]
+        data: typing.Union[str, typing.Dict[str, str]]
     ) -> None:
 
         self.clear()
@@ -101,25 +71,40 @@ class InterfaceProp(dict):
     def add(
         self,
         jail_if: str,
-        bridges: typing.Optional[typing.Union[str, typing.List[str]]]=None,
-        notify: bool=True
+        bridge_if: typing.Optional[str]=None,
+        secure: typing.Optional[bool]=False,
+        notify: typing.Optional[bool]=True
     ) -> None:
+        """
+        add an interface/bridge configuration
 
-        if bridges is None or bridges == [] or bridges == "":
-            return
+        Args:
 
-        if isinstance(bridges, str):
-            bridges = [bridges]
+            jail_if (string):
+                Interface name inside the jail
+
+            bridge_if (string): (optional)
+                Interface name of the host bridge device (VNET only)
+
+            secure (bool): (default=False)
+                Enabling this option adds an epair device between the bridge
+                and the jail, so that source IP and mac address cannot be
+                manipulated from inside the jail
+
+            notify (bool): (default=True)
+                Sends an update notification to the jail config instance
+        """
 
         try:
-            prop = dict.__getitem__(self, jail_if)
-        except KeyError:
-            prop = self.__empty_prop(jail_if)
+            bridge_if = iocage.lib.helpers.parse_none(bridge_if)
+        except:
+            pass
 
-        for bridge_if in bridges:
-            prop.add(bridge_if, notify=False)
+        # ToDo: Implement secur option
 
-        if notify:
+        dict.__setitem__(self, jail_if, bridge_if)
+
+        if notify is True:
             self.__notify()
 
     def __setitem__(self, key, values):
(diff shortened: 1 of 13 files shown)


iocage/cli/start.py
```diff
@@ -98,6 +98,7 @@ def start_jails(jails, logger, print_function):
             jail.require_jail_not_template()
             print_function(jail.start())
         except iocage.lib.errors.IocageException:
+            raise
```
igalic: i suppose this is for debugging purposes.

iocage/lib/Config/Jail/Properties/Interfaces.py
```diff
@@ -24,36 +24,7 @@
 import typing
 
 import iocage.lib.helpers
-
-
-class BridgeSet(set):
```
igalic: i would love to read a bit more (than just "sinularize vnet bridges") of an explanation for removing this whole class

it might also aid a lot in understanding the rest of the code(:

iocage/lib/Config/Jail/Properties/Interfaces.py
```diff
@@ -101,25 +72,36 @@ def set(
     def add(
         self,
         jail_if: str,
-        bridges: typing.Optional[typing.Union[str, typing.List[str]]]=None,
-        notify: bool=True
+        bridge_if: typing.Optional[str]=None,
+        notify: typing.Optional[bool]=True
```
igalic: why is this `Optional`? do we do anything else when it's `None` than when it's `False`?

iocage/lib/Network.py
```diff
@@ -21,75 +21,106 @@
...
-                    logger=self.logger
-                )
+    def setup(self) -> None:
+        if (self.vnet is True) and (self.bridge is None):
+            raise iocage.lib.errors.VnetBridgeMissing(logger=self.logger)
+        self.__create_vnet_iface()
+
+    def teardown(self) -> None:
+        if self.vnet is True:
+            self.__down_host_interface()
```
igalic: is this… dangerous?

iocage/lib/Network.py
```diff
@@ -21,75 +21,106 @@
...
+            if self.bridge.secure is True:
+                self.__down_secure_mode_devices()
+
+    def __down_host_interface(self) -> None:
```
igalic: regardless if it is or isn't, this name is confusing.

igalic: do we have a standard in our code for how to name functions?

`__verb_object_subject(object, ...)`?
`__verb_subject_object(subject, ...)`?
`__verb_object_object(subject, …)?`

is this all confusing enough?

iocage/lib/Network.py
```diff
@@ -21,75 +21,106 @@
...
-        epair_b = f"{epair_a[:-1]}b"
+    def __create_new_epair_interface(self) -> typing.Tuple[str, str]:
+        epair_a = iocage.lib.NetworkInterface.NetworkInterface(
+            name="epair",
+            create=True,
+            group=self.nic_group_name,
+            logger=self.logger
+        )
+        epair_a_name = epair_a.name
+        epair_b_name = f"{epair_a_name[:-1]}b"
```
igalic: what does this mean?

iocage/lib/NetworkInterface.py
```diff
@@ -21,32 +21,40 @@
...
-                 addm=None,
-                 vnet=None,
-                 jail=None,
-                 extra_settings=[],
-                 auto_apply=True,
-                 logger=None):
+    def __init__(
+        self,
+        name: typing.Optional[str]="vnet0",
+        create: typing.Optional[bool]=False,
```
igalic: again, `Optional[bool]` — what's the difference between `False` and `None`?

iocage/lib/NetworkInterface.py
```diff
@@ -99,28 +120,39 @@ def apply_settings(self):
...
             self.exec(command)
 
             if (ipv6 is True) and (address.lower() == "accept_rtadv"):
                 self.exec([self.rtsold_command, self.name])
 
-    def exec(self, command, force_local=False):
+    def exec(
+        self,
+        command: typing.List[str],
+        force_local: typing.Optional[bool]=False
```
igalic: What's the difference between `False` and `None`?

iocage/lib/Config/Jail/Properties/Interfaces.py
gronke: This is a bit awkward. In the first place a device could have multiple bridges it gets assigned to - that does not make sense at all. We never used multiple bridges, but used BridgeSet to store a single item.

iocage/lib/Config/Jail/Properties/Interfaces.py
gronke: You do not need to specify it - it has a default. So it's okay if you call the function with having notify set to None. That's why the interface says it's optional.

iocage/lib/Network.py
gronke: The host interface in this case is the epair half on the host side. The target bridge is never subject of being destroyed 😅 A better method name can help to point this out!

iocage/lib/Network.py
gronke: No standard. It's a private function - it should be easy to read in this files context.

Interesting idea to standardize naming of transformation functions like this. We should make them idempotent too 🤓

iocage/lib/Network.py
gronke: Remove the last character and add a `b` instead. See the [epair man page](https://www.freebsd.org/cgi/man.cgi?query=epair&sektion=4&manpath=FreeBSD+11.1-RELEASE).

iocage/lib/NetworkInterface.py
gronke: Nothing. If the value is `None`, it becomes `False`.

iocage/lib/Config/Jail/Properties/Interfaces.py
```diff
@@ -101,25 +72,36 @@ def set(
...
+
+            jail_if (string):
+                Interface name inside the jail
 
-        if isinstance(bridges, str):
-            bridges = [bridges]
+            bridge_if (string): (optional)
+                Interface name of the host bridge device (VNET only)
 
+                A name beginning with ! (exclamation mark) enables the secure
```
igalic: this is wrong☹

iocage/lib/Config/Jail/Properties/Interfaces.py
igalic: python has named parameters
so their position doesn't really matter (except for `self`)
it makes no sense to type a boolean parameter with `Optional` unless the state of not being set is explicitly distinct to the state of being set.

igalic:
a user can only have one pending review… whenever i forget how much i dislike github's review system, it reminds me again.

iocage/lib/Jail.py
```diff
@@ -1035,15 +1035,23 @@ def _configure_route(self, gateway, ipv6=False) -> None:
 
         ip_version = 4 + 2 * (ipv6 is True)
 
+        # router@interface syntax for static pointopoint route
+        if "@" in gateway:
+            gateway, nic = gateway.split("@", maxsplit=1)
```
igalic: this, too, is new syntax

igalic:
in my very humble opinion, we should make `vnet=yes` default to secure VNETs
if someone wants insecure VNETs, they can set `vnet=insecure`

gronke:
> like, i get that we need them — we probably need both… in one way or another… but i'd rather we didn't introduce two different syntaxes.

I agree, the static routes can be broken in another pull-request.

gronke:
@igalic The two syntaxes are both required to solve the same issue of having multiple public IPs on a bridge that different jails consume.

gronke:
@foo2342 Could you please have a look and see if you can steal a jails IP or Mac address? Denial of service through poisoning the hosts arp table is still possible, but you should not see traffic on the spoofing jail interface.

gronke:
> in my very humble opinion, we should make vnet=yes default to secure VNETs
if someone wants insecure VNETs, they can set vnet=insecure

It should be possible to mix behaviors between jails, so the standard behavior should not change. Instead users can use the newly introduced syntax `interfaces="nic::bridge"` to supply a jail with filtered networking.

PR #228, opened 2017-11-26
make ioc start/stop --rc idempotent

### LINKED ISSUE #227: ioc service's `onestart` should be idempotent, or provide (one)status
our `rc.d` script currently doesn't have a way to retrieve the status, `onestart` is the best way an automation system has to "fire and forget" start it, unfortunately, when jails are already started the result will be:

```
Debug: Service[ioc](provider=freebsd): Could not find ioc in /etc/rc.d
Debug: Executing: '/usr/local/etc/rc.d/ioc onestatus'
Debug: Executing: '/usr/local/etc/rc.d/ioc rcvar'
Debug: Service[ioc](provider=freebsd): rcvar value is YES
Debug: Service[ioc](provider=freebsd): Is enabled
Debug: Executing: '/usr/local/etc/rc.d/ioc onestart'
Error: Could not start Service[ioc]: Execution of '/usr/local/etc/rc.d/ioc onestart' returned 1: * [I|O|C] starting jails... 
Jail 'sealas' is already running
Error: /Stage[main]/Jail::Setup::Libiocage/Service[ioc]/ensure: change from 'stopped' to 'running' failed: Could not start Service[ioc]: Execution of '/usr/local/etc/rc.d/ioc onestart' returned 1: * [I|O|C] starting jails... 
Jail 'sealas' is already running
Debug: Class[Jail::Setup::Libiocage]: Resource is being skipped, unscheduling all events
```

so, we should either make `start` idempotent, or provide a `status`.

igalic:
only filter the jails that aren't running already,
and don't fail when there aren't any.

simple, huh?! This fixes #227!

### DIFF
--- a/iocage/cli/start.py
+++ b/iocage/cli/start.py
@@ -71,9 +71,9 @@ def autostart(
     ]
 ) -> None:
 
-    filters = ("boot=yes", "running=no",)
+    filters = ("boot=yes", "running=no", "template=no",)
 
-    ioc_jails = iocage.lib.Jails.JailsGenerator(
+    ioc_jails = iocage.lib.Jails.Jails(
         logger=logger,
         filters=filters
     )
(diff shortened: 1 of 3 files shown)


iocage/cli/start.py
```diff
@@ -106,8 +106,3 @@ def start_jails(jails, logger, print_function):
 
     if len(failed_jails) > 0:
         exit(1)
-
-    if len(changed_jails) == 0:
-        jails_input = " ".join(list(jails))
-        logger.error(f"No jails matched your input: {jails_input}")
```
gronke: Human users do not recognize the exit code. We need to print an error message on each failure (exit 1).

iocage/cli/start.py
```diff
@@ -106,8 +106,3 @@ def start_jails(jails, logger, print_function):
...
-        exit(1)
```
gronke: During autostart no matching jails should never exit with 1.

iocage/cli/start.py
igalic: why do you consider it a failure?

gronke: You need to input a selector. If no jails match to this, your input is invalid and you want to know about it. Not?

igalic: no, i really don't :P
`ioc start --rc` has a very narrow usecase
if you do it more than once, you (human) will notice a difference
if an automatism does it over and over and over again (because we have no `status`), we really don't care about "errors"

gronke: I see two ways we could:

1.  Silence the output on autostart
2. Duplicate the loop in the autostart function and simplify the code

gronke:
I like your approach to autostart only non-running jails. Whenever this happens the exit code should be 0, even if no jails matched the input.

We should apply the same assumptions to `ioc stop --rc` btw.

igalic:
i'm actually not too fond of this change

there's no differentiation between: my selector is "bad" and my jails didn't start

PR #238, opened 2017-12-28
Fix HardenedBSD detection when using CURRENT release

gronke:
- When using a [custom kernel with CURRENT release](https://github.com/iocage/libiocage#custom-release-eg-running--current) on HardenedBSD the release version detection failed.
- The HardenedBSD uses HTTPS (see https://twitter.com/HardenedBSD/status/937477148503404544) 💜


gronke:
@igalic would you mind to manually test this as well? There were some changes that could affect OS detection on FreeBSD or HardenedBSD. This branch was developed and tested on HBSD-12-CURRENT.

PR #254, opened 2018-02-03
Feature: Updates

gronke:
Not only releases need to be updated, but also jails that were forked from them. The tricky part is that `hbsd-update` and `freebsd-update` behave differently and turned out to work with undocumented flags.

This branch first of all brings support to temporary launch a jail for a single command. This can be done with configuration overrides which are required to temporary alter a jails securelevel.

## Summary
- IocageEvent can be a generator when a rollback step yields. It can either be received as list calling `event.fail()` or consumed directly as generator via `event.fail_generator()`. It will only yield itself if it does not yield nested events.
- CLI command `ioc update <JAIL>` applies the update that was fetched via `ioc fetch -r <RELEASE>`.
- Event rollback steps can optionally be generators
- FreeBSD and HardenedBSD updates can be fetched for releases and jails can be updated independently from the downloaded release updates (no network required)
- Jail/Release updates are executed with Jail.fork_exec(), so that securelevel 0 and other permits can temporarily be set


iocage/cli/update.py
```diff
@@ -0,0 +1,135 @@
...
+    print_function = ctx.parent.print_events
+
+    if len(jails) == 0:
+        logger.error("No jail selector provided")
+        exit(1)
+
+    zfs = iocage.lib.ZFS.ZFS(logger=logger)
+    host = iocage.lib.Host.Host(logger=logger, zfs=zfs)
+
+    if host.distribution.name == "HardenedBSD":
```
igalic: i'm a bit confused what the difference between those two decisions here is

iocage/lib/Jail.py
```diff
@@ -372,6 +372,67 @@ def start(
...
+        if self.config["basejail"] is False:
+            return None
+
+        if self.config["basejail_type"] == "nullfs":
+            return iocage.lib.NullFSBasejailStorage.NullFSBasejailStorage
+
+        if self.config["basejail_type"] == "zfs":
+            return iocage.lib.ZFSBasejailStorage.ZFSBasejailStorage
+
+        return None
```
igalic: i wonder how often we have this decision tree, and in how many places we've forgotten about one of the options

iocage/lib/LaunchableResource.py
```diff
@@ -21,13 +21,14 @@
...
 # POSSIBILITY OF SUCH DAMAGE.
-"""The main CLI for ioc."""
 import libzfs
 import typing
 
 import iocage.lib.Config.Jail.File.RCConf
 import iocage.lib.Config.Jail.File.SysctlConf
 import iocage.lib.Resource
+import iocage.lib.LaunchableResourceUpdate
+UpdaterModule = iocage.lib.LaunchableResourceUpdate
```
igalic: you finally did it, you blew thru the 79 char limit with a class name

iocage/lib/LaunchableResourceUpdate.py
```diff
@@ -0,0 +1,405 @@
...
+        executeReleaseUpdateEvent = events.ExecuteReleaseUpdate(self.release)
+
+        try:
+            for event in jail.fork_exec(self._update_command):
+                yield event
+                self.logger.debug(
+                    f"Update of release '{self.release.name}' finished"
+                )
+        except Exception:
+            raise
```
igalic: this raise without arguments probably needs to go, in favour of what we already have in the next line

iocage/lib/LaunchableResourceUpdate.py
```diff
@@ -0,0 +1,405 @@
...
+                logger=self.logger
+            )
+
+        yield executeReleaseUpdateEvent.end()
+
+        self.logger.verbose(f"Release '{self.release.name}' updated")
+        yield True  # ToDo: yield False if nothing was updated
+
+
+class LaunchableResourceUpdateHardenedBSD(LaunchableResourceUpdate):
```
igalic: this class name makes me uncomfortable 😅

igalic: i think it's mostly because it's Hungarian notation + a Class Hierarchy that already is based on Hungarian notation… and already lives in separate namespaces… so, it seems very pointless and should simply be named
`UpdateHardenedBSD`

iocage/lib/LaunchableResource.py
gronke: 🤣

iocage/lib/Jail.py.rej
```diff
@@ -0,0 +1,53 @@
+***************
```
igalic: i don't think this file was supposed to be committed.

iocage/cli/__init__.py
```diff
@@ -61,10 +61,16 @@
     exit(1)
 
 
-def print_events(generator: typing.Generator[IocageEvent, None, None]) -> None:
+def print_events(
+    generator: typing.Generator[typing.Union[IocageEvent, bool], None, None]
```
igalic: what are the `None`s?
and is `bool` being a terminator a convention, or something you made up?

iocage/lib/Jail.py
```diff
@@ -372,6 +372,89 @@ def start(
 
         self._run_hook("poststart")
 
+    def fork_exec(
```
igalic: is this now public API?

iocage/lib/Jail.py
```diff
@@ -372,6 +372,89 @@ def start(
...
+        self,
+        command: typing.List[str],
+        error_handler: typing.Optional[typing.Callable[
+            [typing.Tuple[subprocess.Popen, str, str]],
+            typing.Tuple[bool, str],
+        ]]=None,
+        **temporary_config_override
+    ) -> typing.Generator['iocage.lib.events.IocageEvent', None, None]:
+        """
+        Starts a jail, runs a command and exists
```
igalic: perhaps explain why the jail needs to be stopped before using this function…
also: does the function stop the jail at the end?

iocage/lib/Jail.py
```diff
@@ -372,6 +372,89 @@ def start(
...
+                err = iocage.lib.errors.CommandFailure(
+                    returncode=child.returncode
+                )
+                yield jailExecEvent.fail(err)
+                raise err
+            else:
+                yield jailExecEvent.skip(err)
+                self.logger.debug(message)
+        else:
+            yield jailExecEvent.end()
```
igalic: does `yield` leave this function? (i.e.: can we swap the `if`/`else` so that we unindent the entire current `if`-block.

gronke:
@igalic your wishes have been heard. Hope you like the new interface names better than the old ones. Code checks pass and I've tested on HardenedBSD and FreeBSD 11.

PR #256, opened 2018-02-19
Move installation of rc script to makefile

worr:
Hey there!

I'm trying to develop an application using libiocage, and I noticed that it currently doesn't install cleanly in a virtualenv, since `setup.py` tries to install an rcscript.

I'm suggesting a change where we install the rcscript as part of the `Makefile`, so that developers can develop programs inside virtualenvs without having to mess around with popping rcscripts in their dev boxes. I'm not wedded to this implementation, and am happy to make changes.

Thanks!!

- [x] Explain the feature
- [x] Read [CONTRIBUTING.md](https://github.com/iocage/iocage/blob/master/CONTRIBUTING.md)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,9 +8,19 @@ deps:
 	python3.6 -m pip install -Ur requirements.txt
 install: deps
 	python3.6 -m pip install -U .
+	@if [ -f /usr/local/etc/init.d ]; then \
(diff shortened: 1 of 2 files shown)


gronke:
@worr thank you, that makes a lot sense. 💛

PR #261, opened 2018-02-23
CLI Fork Exec

gronke:
- Exposes Jail.fork_exec() via the CLI.
- Attach arbitrary data to IocageEvent on message update

### DIFF
--- a/iocage/lib/Jail.py
+++ b/iocage/lib/Jail.py
@@ -378,6 +378,7 @@ class JailGenerator(JailResource):
             [subprocess.Popen, str, str],
             typing.Tuple[bool, str],
         ]]=None,
+        passthru: bool=False,
         **temporary_config_override
     ) -> typing.Generator['iocage.lib.events.IocageEvent', None, None]:
         """Start a jail, run a command and shut it down immediately."""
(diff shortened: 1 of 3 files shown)


daemon.py
```diff
@@ -0,0 +1,34 @@
...
+import http.server
+import urllib.parse
+
+jail = iocage.Jail('foo')
+
+class ForkExecDaemon(http.server.BaseHTTPRequestHandler):
+
+    def do_GET(self):
+        
+        command = urllib.parse.unquote(self.path.strip("/")).strip(" ").split(" ")
```
gronke: Bandit and @igalic will love this!

iocage/lib/events.py
```diff
@@ -176,8 +176,11 @@ def duration(self) -> typing.Optional[float]:
             return None
 
     def _update_message(self, **kwargs) -> None:
-        if "message" in kwargs:
-            self.message = kwargs["message"]
+        for key in list(kwargs):
```
skarekrow: This could be `for key in kwargs.keys:` no?

gronke:
### On the server
```
$ ioc create -b -n fork-exec-test exec_start="/usr/bin/true"               
fork-exec-test successfully created from 11.1-RELEASE!
$ python3.6 experiments/socket-based-daemon.py fork-exec-test
whoami && hostname
127.0.0.1 - - [01/Jan/1970 13:37:00] "POST / HTTP/1.1" 200 -
```

### On the client
```
$ /usr/bin/time -l curl -i --data "whoami && hostname" 127.0.0.1:8080/
HTTP/1.0 200 OK
Server: BaseHTTP/0.6 Python/3.6.4
Date: Sat, 24 Feb 2018 13:16:14 GMT
Content-type: text/plain

root
fork-exec-test
0.30 real         0.00 user         0.00 sys
      6480  maximum resident set size
       192  average shared memory size
         8  average unshared data size
       128  average unshared stack size
       317  page reclaims
         0  page faults
         0  swaps
         3  block input operations
         0  block output operations
         1  messages sent
         2  messages received
         0  signals received
         6  voluntary context switches
         3  involuntary context switches
```

iocage/lib/events.py
```diff
@@ -176,8 +176,11 @@ def duration(self) -> typing.Optional[float]:
             return None
 
     def _update_message(self, **kwargs) -> None:
-        if "message" in kwargs:
-            self.message = kwargs["message"]
+        for key in kwargs.keys():
+            if key == "message":
+                self.message = kwargs["message"]
+            else:
+                self.data[key] = kwargs[key]
```
gronke: This allows attaching data to events. In the case of events JailExec we can read stdout and stderr from it when the event has ended.

iocage/lib/Jail.py
```diff
@@ -417,27 +418,41 @@ def _revert_jail_started():
...
         jailExecEvent.add_rollback_step(_revert_jail_started)
 
-        child, stdout, stderr = self.exec(command, ignore_error=True)
-        if child.returncode != 0:
-            if (error_handler is None):
-                passed = False
-                message = ""
+        if passthru is True:
+            try:
+                self.passthru(command)
```
gronke: Passthru enables CLI users to interact with an executed command.

igalic: given how long and unreadable this if/else is, maybe we should also bundle the else's call into a separate function like `passthru`

gronke:
@igalic I've tried refactoring the generative exec approach into another method. Because the steps tightly coupled to the events initialized in the fork_exec method, I think it is not an improvement to break out another private method here.

PR #265, opened 2018-03-01
Feature: Clone

### LINKED ISSUE #210: CLI: clone/promote
- clone a jail
- promote the cloned jail

gronke:
closes #210

Introduces the `ioc clone` CLI command:
```
ioc clone <FROM_JAIL> <TO_JAIL>
ioc promote <TO_JAIL>
ioc destroy <FROM_JAIL>
```

- Recursive cloning of datasets / snapshots
- Export jails to a dataset
- Jails can be cloned from existing ones
- Cloned jails can be promoted (so that the parent can be deleted)


iocage/lib/ZFS.py
```diff
@@ -24,16 +24,31 @@
...
 import iocage.lib.errors
 
+Logger = iocage.lib.Logger.Logger
+
 
 class ZFS(libzfs.ZFS):
     """libzfs enhancement module."""
 
-    logger: typing.Optional[iocage.lib.Logger.Logger] = None
+    _logger: typing.Optional['iocage.lib.Logger.Logger']
```
gronke: This change is required for compliance with the typing. The logger on ZFS is optional and is not set when initializing the object, but can be set at any later time (using the setter method).

gronke:
In an offline discussion @yonk42 mentioned that a parent of a cloned jail cannot be deleted unless they get promoted. This is true and needs to be addressed, but I'm unsure whether automatically promoting could be an option. If we do not automatically promote another jail on deletion of the parent, further user interaction is required.

gronke:
@foo2342 brought up an interesting point while discussing the problem. Instead of automatically promoting the next available sibling the former parent could be moved in an archive dataset (instead of `iocage/jails`), so that the datasets persists until the last clone was deleted without getting recognized as startable jail. It is reasonable to delete every snapshot that was not cloned by another jail on parent jail deletion.

iocage/lib/ZFS.py
igalic: i think it might be time that we (once again) rethink our logger (initialization (method))

gronke: This is a special occasion. Having a logger in this class available improves our output, but the ZFS class itself should not depend on having a logger. (Imagine you want to use the module in another context than libiocage). I also remember an issue overriding the default ZFS __init__ method.

PR #268, opened 2018-03-02
Strong Typing

gronke:
Another attempt to enforce typed interfaces without getting eaten by merge conflicts.


.travis/mypy-stubs/libzfs.pyi
```diff
@@ -336,53 +336,3 @@ class ZfsConverter:
     __init__ = ... # type: Any
     to_native = ... # type: Any
     to_property = ... # type: Any
-
-class datetime(date):
```
igalic: oh?

gronke: It was never used in our context, but caused confusion. That's why I have removed it. Do you think that was an error?

iocage/lib/Releases.py
```diff
@@ -31,19 +31,23 @@
 
 # MyPy
 import libzfs
+ReleaseListType = typing.List[iocage.lib.Release.ReleaseGenerator]
```
igalic: more of this please!

iocage/lib/Config/Jail/BaseConfig.py
```diff
@@ -146,11 +146,22 @@ def update_special_property(self, name: str) -> bool:
...
+        )
+    ) -> None:
         """Attach a special property to the configuration."""
         self.special_properties[name] = special_property
 
-    def _set_legacy(self, value, **kwargs) -> None:
+    def _set_legacy(  # noqa: T484
+        self,
+        value: typing.Union[bool, str],
+        **kwargs
```
igalic: please open up an issue for cleaning up `noqa: T484`

gronke: @igalic as you wish #282

PR #272, opened 2018-03-18
Add JID to IOCAGE_ env variables

urosgruber:
This fix adds JID to the list of ENV variables. It's set to None in case of **prestart** and to JID in case of **poststart**


gronke:
No idea why it wasn't there in first place 😄 Thanks for the PR!

PR #294, opened 2018-04-07
Feature: import/export

### LINKED ISSUE #208: CLI: import/export
- Export jails to archive
- Import jails from archive

gronke:
closes #208


gronke:
A jails fstab file needs to be preserved too. When exporting it the paths in the fstab file should be updated to a provisional path that is adapted to the real jail name when importing.

PR #338, opened 2018-05-06
Load jail ZFS shares before exec.start

### LINKED ISSUE #337: Mount ZFS shares during jail prestart
The jail ZFS shares are currently mounted after the jail command finished. Since we moved to towards initialization during Jail `exec.prestart`, this feature needs to be migrated to a QueuedCommand (like VNET networking), so that shares are available in fork_exec and on service initialization.

gronke:
closes #337 

blocked by https://reviews.freebsd.org/D15330


urosgruber:
https://svnweb.freebsd.org/base?view=revision&revision=337863

iocage/lib/Config/Jail/Defaults.py
```diff
@@ -109,6 +109,7 @@ class JailConfigDefaults(iocage.lib.Config.Jail.BaseConfig.BaseConfig):
         "exec_clean": 1,
         "exec_fib": 1,
         "exec_prestart": None,
+        "exec_started": None,
```
igalic: we should probably rename this now to be in line with https://svnweb.freebsd.org/base?view=revision&revision=337863

iocage/lib/Jail.py
```diff
@@ -408,6 +407,7 @@ def start(
 
         exec_prestart: typing.List[str] = []
         exec_start: typing.List[str] = []
+        exec_started: typing.List[str] = []
```
igalic: we should probably rename this now to be in line with https://svnweb.freebsd.org/base?view=revision&revision=337863

iocage/lib/Jail.py
```diff
@@ -420,12 +420,21 @@ def start(
...
             exec_poststart += _poststart
 
+        if self.config["jail_zfs"] is True:
+            share_storage = iocage.lib.ZFSShareStorage.QueuingZFSShareStorage(
+                jail=self,
+                logger=self.logger
+            )
+            share_storage.mount_zfs_shares()
+            exec_start += share_storage.read_commands("jail")
+            exec_started += share_storage.read_commands()
```
igalic: same as above ;)

PR #344, opened 2018-05-09
re-apply jail command environment variables

### LINKED ISSUE #342: Hook scripts missing IOCAGE_ env variables
I'm not sure if feature was changed but I noticed new folder **launch-scripts** including some shell scripts and all are also exposed inside jail as well. I can find any info of how this can be used or what is the case for this. But after some testing again with latest changes I noticed there is no `IOCAGE_` evn variables present inside pre/post-start and pre/post-stop scripts.

Am I missing something or configuration changed. I usually set this scripts via
`ioc set exec_poststart=/root/scripts/_poststart.sh ioc/myjail`, Also previously I was able to echo progress but latest version does not output anything on the screen.

gronke:
fixes #342 

- disables jail hook execution with exec.clean by default
- passes a selected list of environment variables to the jail start command


urosgruber:
Is this partial fix until `jail` is fixed? Just asking if it's worth testing this so it can be merged to master.

gronke:
I'm going to chat with @igalic tonight to figure out whether the recent changes that configure networking and ZFS shares via `exec.*` hooks rather than forking individual processes from Python need to be reverted or what our other options are. @urosgruber I'll give you a heads up as soon as it makes sense to do further tests. Sorry for the delay and breaking the feature in first place!

urosgruber:
@gronke no problem. For now, I hooked it up via JID that is still present but would be nice to add this back in. Especially jail path is not available even via `ioc get` so I needed to write some workarounds.

gronke:
@urosgruber would you try this branch again please? All jail configuration variables should be available in any hook now.

The IOCAGE_JID is not yet available in exec.start, but in exec.poststart. This will be addressed in another PR.

urosgruber:
I'll test it today.

urosgruber:
I hope I was running proper version as previously jails could not be destroyed with this version (did you rebased with master so that fix for this is already included). 

Otherwise ENV variables are identical to those libiocage version from about a month ago. I think it's safe to merge it.

gronke:
Thanks for your reviews @foo2342, @urosgruber and @igalic!

urosgruber:
@gronke I just tested this feature once again and I'm not sure if something was changed again where feature does not work. I have exact same scripts for prestart and poststart but none output any IOCAGE_ env variables

PR #345, opened 2018-05-09
restore support for multiple vnet devices

### LINKED ISSUE #343: Cannot start jails with multiple VNET interfaces
It is not possible to attach multiple VNET interfaces using the jail argument `vnet.interface`. The right spot to manually jail the epair device is not available yet.

The [jail command startup order](https://github.com/freebsd/freebsd/blob/f2a9371551afae9862cf771255b1f2d2704cced7/usr.sbin/jail/jail.c#L86-L106) defines no hook on host side after jail creation but before `exec.start` (within the jail).

Blocked by https://reviews.freebsd.org/D15330

gronke:
fixes #343 

Delays execution of `exec.start` to `exec.poststart`. By doing that we already know the jails JID, so that we can manually assign a network device to it. `vnet.interface` is doing the same, but can only deal with a single network device.


iocage/lib/Jail.py
```diff
@@ -501,6 +520,25 @@ def _run_poststop_hook_manually(self) -> None:
...
+        """Wrap a jail hook command for a host hook script."""
+        if commands is None:
+            return []
+
+        EOF_IDENTIFIER = f"EOF{random.getrandbits(64)}"
+        output: typing.List[str] = [
+            "set -e",
+            "echo 'Executing jail start scripts'",
+            "jexec -j {self.identifier} /bin/sh <<{EOF_IDENTIFIER}"
+        ] + commands + [
```
igalic: does `+` automatically expand as `commands.join("\n")`?

iocage/lib/Jail.py
```diff
@@ -426,6 +426,25 @@ def start(
         if self.config["exec_start"] is not None and (single_command is None):
             exec_start += [self.config["exec_start"]]
 
+        if self._has_multiple_interfaces:
+            self.logger.debug(
+                "Running exec.start in multiple VNET compatibility mode"
+            )
+            # the jail command misses the ability to run host commands when the
```
igalic: `misses` → `is missing`

iocage/lib/Network.py
```diff
@@ -175,72 +164,62 @@ def __create_new_epair_interface(
...
         nic_suffix_b: str=":b",
         **nic_args: typing.Any
-    ) -> PrestartCommandList:
+    ) -> StartedCommandList:
 
-        commands_start: PrestartCommandList = []
+        commands_started: StartedCommandList = []
 
-        commands_start += iocage.lib.NetworkInterface.QueuingNetworkInterface(
+        commands_started += iocage.lib.NetworkInterface.QueuingNetworkInterface(
```
igalic: `./iocage/lib/Network.py:174:80: E501 line too long (80 > 79 characters)`

iocage/lib/Jail.py
gronke: This concatenates multiple lists of string (such as `commands`) that are later on joined.

Python joins are the other way around: `"\n".join(commands)` by the way.

PR #352, opened 2018-05-20
fix vnet_interface config merging

gronke:
- The jail configuration property `vnet_interfaces` is a complex type (list) that needs to be wrapped accordingly
- `exec_clean` is a boolean property

### DIFF
--- a/iocage/lib/Config/Jail/BaseConfig.py
+++ b/iocage/lib/Config/Jail/BaseConfig.py
@@ -271,6 +271,33 @@ class BaseConfig(dict):
     def _has_legacy_tag(self) -> bool:
         return "tag" in self.data.keys()
 
+    def _get_vnet_interfaces(self) -> typing.List[str]:
+        return list(
+            iocage.lib.helpers.parse_list(self.data["vnet_interfaces"])
+        )
+
+    def _set_vnet_interfaces(  # noqa: T484
+        self,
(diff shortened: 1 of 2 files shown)


iocage/lib/Config/Jail/BaseConfig.py
```diff
@@ -271,6 +271,33 @@ def _set_tag(self, value: str, **kwargs) -> None:  # noqa: T484
...
+        return list(
+            iocage.lib.helpers.parse_list(self.data["vnet_interfaces"])
+        )
+
+    def _set_vnet_interfaces(  # noqa: T484
+        self,
+        value: typing.Union[str, typing.List[str]],
+        **kwargs
+    ) -> None:
+        if isinstance(value, str) is True:
```
igalic: i miss erlang (or elixir (or rust…))

gronke:
Let me make an example:
```python3
import iocage
jail = iocage.Jail("myjail")
jail.config["vnet_interfaces"] = ["epair1b", "epair2b"]
jail.config["exec_clean"] = False
jail.config["exec_clean"] = "no"
# failure
jail.config["exec_clean"] = "wrong-input"
``` 

Before this changes you would have to always use string values. Even invalid values would have been possible, since we only want values that represent a boolean to be stored in the jail configuration. (Possible values are `yes`, True, `true`, etc - even though we store them in a normalized form)

PR #360, opened 2018-05-23
Fix release fetching and idempotence on FreeBSD

gronke:
- Attach jail launch command stdout to JailLaunch event on end
- Test update command output for `No updates are available to install.` on FreeBSD and mark the event as skipped
- Fixes an issue where the JailForkExec event hangs when the process did not print to stdout

### DIFF
--- a/iocage/lib/ResourceUpdater.py
+++ b/iocage/lib/ResourceUpdater.py
@@ -488,6 +488,7 @@ class FreeBSD(Updater):
         _command = "\n".join([
             "set +e",
             f"OUTPUT=\"$({command})\"",
+            "echo $OUTPUT",
             "RC=$?",
             "if [ $RC -gt 0 ]; then",
             (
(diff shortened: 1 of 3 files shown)


iocage/lib/ResourceUpdater.py
```diff
@@ -488,6 +496,7 @@ def _wrap_command(self, command: str) -> str:
         _command = "\n".join([
             "set +e",
             f"OUTPUT=\"$({command})\"",
+            "echo $OUTPUT",
```
igalic: Was this part of the debugging code, or is it necessary / useful?

gronke: The exit code of some `freebsd-update.sh` versions is `1` in case no updates were installed. This wrapper script checks whether this is the case and exits with `0` anyway, so that libiocage does not complain about the failed execution. Later on we read the output again (from iocage.lib.events.JailLaunch.stdout) and determine whether the updates were skipped or some have been newly installed.

PR #362, opened 2018-05-24
Improve Resource Limits

gronke:
- It was not possible to set the `per` part of an rctl rule
- The original iocage declaration `<amount>:<action>` is the opposite of the syntax used by rctl
- The enhanced implementation accepts both styles (e.g. `vmemoryuse=2g:deny` or `vmemoryuse=deny=2g/jail`) and stores the iocage compatible style when possible (as long as no `per` other than `/jail` was used)
- All resource limits are special properties now

### DIFF
--- a/iocage/lib/Config/Jail/BaseConfig.py
+++ b/iocage/lib/Config/Jail/BaseConfig.py
@@ -67,8 +67,8 @@ class BaseConfig(dict):
 
     """
 
-    special_properties: 'iocage.lib.Config.Jail.Properties.Properties'
     data: typing.Dict[str, typing.Any]
+    special_properties: 'iocage.lib.Config.Jail.Properties.Properties'
 
     def __init__(
         self,
@@ -134,22 +134,6 @@ class BaseConfig(dict):
                 del data["uuid"]
 
         self.clone(data)
-
-    def update_special_property(self, name: str) -> bool:
-        """Triggered when a special property was updated."""
-        try:
-            self.data[name] = str(self.special_properties[name])
-            return True
-        except KeyError:
-            return False
-
-    def attach_special_property(
-        self,
-        name: str,
-        special_property: 'iocage.lib.Config.Jail.Properties.Property'
-    ) -> None:
-        """Attach a special property to the configuration."""
-        self.special_properties[name] = special_property
 
     def _set_legacy(  # noqa: T484
         self,
@@ -499,13 +483,14 @@ class BaseConfig(dict):
         except AttributeError:
             return False
 
-    def __getitem_user(self, key: str) -> typing.Any:
+    def _getitem_user(self, key: str) -> typing.Any:
         try:
             # passthrough existing properties
             return self.__getattribute__(key)
         except AttributeError:
             pass
 
+        # special property
         is_special_property = self.special_properties.is_special_property(key)
         is_existing = key in self.data.keys()
         if (is_special_property and is_existing) is True:
@@ -536,7 +521,7 @@ class BaseConfig(dict):
         A KeyError is raised when no criteria applies.
         """
         try:
-            return self.__getitem_user(key)
+            return self._getitem_user(key)
         except KeyError:
             pass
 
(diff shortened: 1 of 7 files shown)


iocage/lib/Jail.py
```diff
@@ -1633,53 +1636,29 @@ def _limit_resources(self) -> None:
...
-            self._resource_limit_config_keys
-        ):
-
-            if (limit is None) and (action is None):
-                # this resource is not limited (limit disabled)
+        for key in iocage.lib.Config.Jail.Properties.ResourceLimit.properties:
+            try:
+                rlimit_prop = self.config[key]
+            except KeyError:
                 continue
```
igalic: logging happens on set? right?
so, when loading a corrupt legacy config, we don't know… correct?

gronke: We do because parsing the rule fails. The next changes will call set to initialize the value anyway, so that it has to pass the parser once.

igalic: sorry, i meant: we don't tell the user

iocage/lib/Config/Jail/JailConfig.py
```diff
@@ -85,18 +85,18 @@ def _get_host_hostname(self) -> str:
                 return str(jail.humanreadable_name)
             raise e
 
-    def __setitem__(  # noqa: T484
+    def __setitem__(
         self,
         key: str,
         value: typing.Any,
-        **kwargs
```
igalic: why was this necessary before?

iocage/lib/Config/Jail/BaseConfig.py
```diff
@@ -114,9 +114,13 @@ def clone(
             if (key in ["id", "name", "uuid"]) and (current_id is not None):
                 value = current_id
 
-            self.__setitem__(key, value, skip_on_error=skip_on_error)
+            self.__setitem__(  # noqa: T484
```
igalic: how come this isn't necessary with other `__setitem__` function(call?)s?

iocage/lib/Config/Jail/BaseConfig.py
```diff
@@ -499,13 +487,14 @@ def _skip_on_error(self, **kwargs) -> bool:  # noqa: T484
...
-    def __getitem_user(self, key: str) -> typing.Any:
+    def _getitem_user(self, key: str) -> typing.Any:
         try:
             # passthrough existing properties
             return self.__getattribute__(key)
         except AttributeError:
             pass
 
+        # special property
         is_special_property = self.special_properties.is_special_property(key)
```
igalic: this looks like some form of tautology…

iocage/lib/Config/Jail/Properties/ResourceLimit.py
```diff
@@ -0,0 +1,192 @@
...
+
+class ResourceLimitProp(ResourceLimitValue):
+    """Special jail config property for resource limits."""
+
+    def __init__(
+        self,
+        config: typing.Optional[
+            'iocage.lib.Config.Jail.BaseConfig.BaseConfig'
+        ]=None,
+        property_name: str="ip4_address",
```
igalic: why is this the default `property_name`?

iocage/lib/Config/Jail/JailConfig.py
gronke: because of the `skip_on_error` property. But since that's the only argument it's better to specify it explicitly instead of `**kwargs`.

iocage/lib/Config/Jail/BaseConfig.py
gronke: This __setitem__ function extends the method with the skip_on_error argument. MyPy thinks it is invalid to do so.

iocage/lib/Config/Jail/Properties/ResourceLimit.py
gronke: There's no need for a default. Will remove.

PR #365, opened 2018-05-26
Update of multiple jails

gronke:
- Prevent events from behaving as singleton
- Use the resource as identifier of update events

The event identifier of ReleaseUpdate events was the releases full_name. When updating resources that had the same event, the identifier led to confusion in the event history stack.

### DIFF
--- a/iocage/lib/ResourceUpdater.py
+++ b/iocage/lib/ResourceUpdater.py
@@ -285,10 +285,10 @@ class Updater:
             f"{dataset.name}@pre-update"
         )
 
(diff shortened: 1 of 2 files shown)


gronke:
This can be tested by running:
```sh
ioc fetch -r 11.1-RELEASE
ioc create -r 11.1-RELEASE -b -n jail-a
ioc create -r 11.1-RELEASE -b -n jail-b
ioc update *
```

igalic:
i was not aware of the `+` syntax.
we should document this behaviour in #374

Issue #373, opened 2018-05-28
ioc start '*' failing when some jails already running


igalic:
currently `ioc start '*'` fails if some of the jails are already running.
conversely, `ioc stop '*'` fails if some of the jails are already stopped.

Should we continue this behaviour?
***specifically*** when the filter is `*`.

gronke:
Personally I find is more consistent to always target the same jails in an operation. Pre-defining a `running=yes` filter changes the behavior from "Will apply the start operation on all matching jails" to "Will do some magic to enhance your input". Imagine we want to do the same with the set operation. Should it also only address stopped jails and not complain at all when a property wasn't set because the jail was never selected.

igalic:
aye — thanks for the explanation

Issue #372, opened 2018-05-28
ioc stop '*' should not select templates


igalic:
```
# ioc stop '*'
Jail 'tpl-db' is not running
[+] JailDestroy@web01: OK [0.179s]: 1
web01 stopped
[+] JailDestroy@db01: OK [0.199s]: 1
db01 stopped
[+] JailDestroy@web02: OK [0.182s]: 1
web02 stopped
Jail 'tpl-web' is not running
[+] JailDestroy@db02: OK [0.177s]: 1
db02 stopped
*** Error code 1
```

of course they are *not* running
They can never run. We defined them such that they could never possibly run.
`ioc stop` should not select jails with `template=yes` for its actions, as `ioc start` ***already does***!

Issue #371, opened 2018-05-28
consider validating ip4 / ip6 addresses on set


igalic:
currently when creating a jail, we can currently create it with a perfectly invalid IP:

```
# ioc create --basejail --name tpl-web ip4_addr="vtnet0|.2/16"
tpl-web successfully created from 11.1-RELEASE!
```

only on start can we see that something may be off:

```
# ioc start tpl-web
[-] JailLaunch@tpl-web: FAILED [0.135s]: 2
Launching jail tpl-web failed
*** Error code 1
```

but not quite what.

we may want to validate this.

igalic:
like how *cough* #113 *cough* i thought the [pathlib module](https://docs.python.org/3/library/pathlib.html) seems interesting…

gronke:
True, both are pretty much made for those tasks. PRs welcome 👍

Issue #361, opened 2018-05-23
UX of setting rlimits


igalic:
currently it's possible to set rlimits, but it has no effect on a default setup

```
rctl: RACCT/RCTL present, but disabled; enable using kern.racct.enable=1 tunable
```

We should check for this `sysctl`, and either set it ourselves, or recommend our users set it themselves.

*****

i don't think we can set this from ioc:

```
sysctl: oid 'kern.racct.enable' is a read only tunable
sysctl: Tunable values are set in /boot/loader.conf
```

so what's left is to recommending to the users they do this themselves.

urosgruber:
I'm ok to just tell the user it's disabled and maybe some help message.

gronke:
There is another occasion where host features could be disabled: Secure-VNET requires ipfw to be enabled with specific features. In both cases it makes sense to do the compatibility checks early on, so that we do not produce jail zombies.

Would you find it useful to include the commands to enable such host features in the message or is the plain warning about missing features enough to figure it out?

urosgruber:
@gronke commands would be great.

gronke:
Implemented and documented in the Wiki https://github.com/iocage/libiocage/wiki/Resource-Limits

urosgruber:
@gronke are you sure resource-limits are implemented the way wiki is describing?

gronke:
> @gronke are you sure resource-limits are implemented the way wiki is describing?

@urosgruber Can you please make an example of what is not working as expected?

urosgruber:
So if I call `ioc set vcpu=vmem:deny=128M/jail foobar` I get an error **The config property 'vcpu' of jail 'foobar' is unknown.**

But `ioc set vmemoryuse=128M foobar` works.

PR #374, opened 2018-05-28
test almost all features we support

igalic:
the following features are tested

* fetch (default + non-default release)
* create a basejail
* execute commands (pkg)
* convert jail into template
* create jails from these templates
* start those jails
* stop them
* cleanup

we test two ways of skipping prompts. `--force`, and `yes |`


igalic:
in creating this test, i've uncovered the following issues:

- #372 
- #373 
- #371

igalic:
What's missing?

- HardenedBSD
- vnet
- jail in jail
- resource limits (#361 #362)

Extra nice-to-haves:

does the fact that we can execute `pkg` prove anything about the jail?
Can we test containment?
Can we test cross-jail-communication?

gronke:
Great idea to quickly get a thumbs up for changes by running this script. I would prefer to put it in a separate file. Maybe `iocage/tests/run-integration.sh`?

iocage/tests/run-integration.sh
```diff
@@ -0,0 +1,62 @@
...
+ioc list --no-header --output-format=list | grep -c web0 | grep -qw 2
+
+# stop all
+ioc stop '*'
+
+# cleanup
+ioc destroy --force '*01'
+yes | ioc destroy '*02'
+# sleep for (u)mounts to settle:
+sleep 2
```
igalic: btw, has anyone noticed this little bundle of joy here?
(i reckon it's `umounts` that need to finish up… the period is fairly short)

iocage/tests/run-integration.sh
```diff
@@ -0,0 +1,62 @@
...
+ioc destroy --force template=yes
```
igalic: it's required because otherwise this fails
and i reckon it might be related to #295 ???

iocage/tests/run-integration.sh
gronke: At the time ioc exists, all operations should be finished. Are you sure a delay is required? If so, we need to create a bug ticket for it.

PR #375, opened 2018-05-29
ignore: pytest_cache/

igalic:
on those rare systems we can actually test on, we don't want
.pytest_cache/ in our git status

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -117,3 +117,4 @@ ENV/
 
 *.orig
 *.rej
+.pytest_cache/


gronke:
If we start ignoring `.vscode/` assets, shouldn't we also include ignore assets from all other editors? You could ignore it globally, so that you won't commit it to any other repository as well. I don't mind merging it like this because it is obvious to be about code editor assets.

igalic:
good point
i think we do ignore most assets of most editors already:

```
# Editor and OS Files
*.swp
.#*
._*
.DS_Store
.idea/
```

PR #386, opened 2018-05-31
fstab cli command ctx.host must be a HostGenerator

gronke:
Fixes the broken fstab CLI command.


gronke:
@igalic 😬

igalic:
oops sorry i overlooked it

gronke:
No need to be sorry. It's just a reminder that we need more test coverage now.

PR #387, opened 2018-05-31
Error handling when launching a jail

gronke:
Attaching a pseudo-terminal to the jail launch command resulted in invalid error detection on jail launch. Failing commands did not influence the jail start command exit code and did not revert the operations accordingly.

### DIFF
--- a/iocage/lib/Jail.py
+++ b/iocage/lib/Jail.py
@@ -1515,12 +1515,6 @@ class JailGenerator(JailResource):
             passthru=passthru
         )
 
-        if returncode > 0:
-            raise iocage.lib.errors.JailCommandFailed(
-                returncode=returncode,
-                logger=None
-            )
-
         return stdout, stderr, returncode
 
     def _get_value(self, key: str) -> str:
--- a/iocage/lib/Jail.py
+++ b/iocage/lib/Jail.py
@@ -410,7 +410,9 @@ class JailGenerator(JailResource):
         exec_prestart: typing.List[str] = []
         exec_start: typing.List[str] = []
         exec_started: typing.List[str] = [
-            f"echo \"export IOCAGE_JID=$IOCAGE_JID\" > {self.script_env_path}"
+            f"echo \"export IOCAGE_JID=$IOCAGE_JID\" > {self.script_env_path}",
+            "set -e",
+            "set -u",
         ]
         exec_poststart: typing.List[str] = []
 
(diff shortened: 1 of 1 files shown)


iocage/lib/Jail.py
```diff
@@ -410,7 +410,9 @@ def start(
         exec_prestart: typing.List[str] = []
         exec_start: typing.List[str] = []
         exec_started: typing.List[str] = [
-            f"echo \"export IOCAGE_JID=$IOCAGE_JID\" > {self.script_env_path}"
+            f"echo \"export IOCAGE_JID=$IOCAGE_JID\" > {self.script_env_path}",
+            "set -e",
+            "set -u",
```
igalic: this can be just one line `set -eu` but we should maybe just have it… uhm…automatically written… elsewhere.

iocage/lib/Jail.py
```diff
@@ -465,8 +467,11 @@ def start(
         self._write_hook_script(
             "poststart",
             self._wrap_hook_script_command_string([
+                "set -u",
+                "set -e",
+                "set -x",
                 "/bin/echo running exec.started hook on the host",
-                f"/bin/sh {self.get_hook_script_path('started')}",
+                f"/bin/sh {self.get_hook_script_path('started')} 2>&1",
```
igalic: can you briefly explain why we cannot get `stdin` & `stdout` at forwarded to us?

iocage/lib/Jail.py
```diff
@@ -506,6 +511,11 @@ def _stop_failed_jail(
                 )
             if stdout is not None:
                 self.logger.spam(stdout)
+            if returncode != 0:
```
igalic: :+1:

iocage/lib/Jail.py
```diff
@@ -1420,10 +1430,7 @@ def _launch_persistent_jail(
             passthru=passthru
         )
         if returncode > 0:
```
igalic: please also change it here.

iocage/lib/Jail.py
gronke: We iterate over stdout to get async access. When stdout is written the execution is stuck because nobody consumes the arriving data. We can solve it with threading, but that can be achieved in a later step. In the meanwhile we need to combine stderr and stdout.

iocage/lib/Jail.py
igalic: we have more of that here:

```
iocage/lib/JailState.py:112:        if returncode > 0:
iocage/lib/JailState.py:136:        if returncode > 0:
iocage/lib/JailState.py:214:        if returncode > 0:
iocage/lib/JailState.py:235:        if returncode > 0:
iocage/lib/Jail.py:1422:        if returncode > 0:
iocage/lib/Jail.py:1515:        if returncode > 0:
iocage/lib/helpers.py:175:    if returncode > 0:
```

i'm not being paranoid here, or?

gronke: Good catch. We evaluate the error codes at the calling functions. The premature exist (still returning the `stdout`, `stderr`, `returncode` tuple) is just to leave the function early.

PR #388, opened 2018-05-31
Fix fetching EOL releases

### LINKED ISSUE #289: fetching updates of EOL releases fails
Fetching updates for an EOL release (like 11.0-RELEASE) fails because the fetch command exits with `1` and the following warning:

```
...
/usr/share/man/man8/ntpd.8.gz
/usr/share/man/man8/ntpdc.8.gz
/usr/share/man/man8/ntpq.8.gz
/usr/share/man/man8/sntp.8.gz
/usr/share/man/mandoc.db
/usr/share/openssl/man/mandoc.db

WARNING: FreeBSD 11.0-RELEASE HAS PASSED ITS END-OF-LIFE DATE.
Any security issues discovered after Thu Nov 30 23:59:59 UTC 2017
will not have been corrected.
```

libiocage should present a warning as well, but accept the command status error when executing the actual fetch internally.

gronke:
fixes #289

When downloading release updates with freebsd-update.sh the script exists with exit code 1 when a release was fetched that is listed as EOL. We have already warned the user once in the release list, so that we can assume the fetching was intentional and has finished successfully after reading this message.

The hbsd-update tool on older HardenedBSD is lacking flags to keep temporary files. HardenedBSD jails with userland version 10.2 and before are unsupported at all.

### DIFF
--- a/iocage/lib/Release.py
+++ b/iocage/lib/Release.py
@@ -407,15 +407,17 @@ class ReleaseGenerator(ReleaseResource):
         return self._parse_release_version(self.name)
 
     def _parse_release_version(self, release_version_string: str) -> float:
-        parsed_version, suffix = release_version_string.split("-", maxsplit=1)
+        _parts = release_version_string.split("-", maxsplit=1)
+        parsed_version = _parts[0]
         try:
             version = float(parsed_version)
             if self.host.distribution.name == "HardenedBSD":
-                has_stable_suffix = (suffix.upper() == "STABLE") is True
-                if (version == 10.0) and has_stable_suffix:
-                    return 10.4
-                elif (version == 11.0) and has_stable_suffix:
-                    return 11.1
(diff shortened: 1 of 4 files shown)


iocage/lib/Release.py
```diff
@@ -367,6 +368,60 @@ def annotated_name(self) -> str:
...
+        _parts = release_version_string.split("-", maxsplit=1)
+        parsed_version = _parts[0]
+        try:
+            version = float(parsed_version)
+            if self.host.distribution.name == "HardenedBSD":
+                if (len(_parts) == 2) and (_parts[1].upper() == "STABLE"):
+                    if (version == 10.0):
+                        return 10.4
+                    elif (version == 11.0):
+                        return 11.1
```
igalic: from what i'm reading here, this function will need constant fixing.

PR #399, opened 2018-06-08
Resource-Limits compatibility with iocage_legacy v1.7.6

gronke:
- Ignore resource limits that were set to `off` or `None`
- Execute rctl from jail hook commands


igalic:
aaah, yes, idempotency…

iocage/lib/Jail.py
```diff
@@ -1684,35 +1681,35 @@ def _configure_nameserver(self) -> None:
...
-            return
+            return []
 
         self.logger.verbose("Clearing resource limits")
-        iocage.lib.helpers.exec(
-            ["/usr/bin/rctl", "-r", f"jail:{self.identifier}"],
-            logger=self.logger,
-            ignore_error=True
-        )
+        return [f"/usr/bin/rctl -r jail:{self.identifier} 2>/dev/null || true"]
```
igalic: is the || true strictly necessary?

gronke: The command may exit with 1 but should not raise an error. The other option is to temporary `set -e`. Any preference?

igalic: *brrrr*

a comment that it may fail and we don't care at this point.

Issue #377, opened 2018-05-29
inconsistency between template=no and template=no,- usage


igalic:
in 8b1fea9b06b44cdedc813c949e7d51176b5906cd @gronke introduced a fix for iocage_legacy flavours which use `template=none` instead of `template=no`

However, many parts of the code still only assume `template=no` when dealing with jails

This should be fixed.

gronke:
The internal storage jail.config.data should always contain a Boolean value. The output to ZFS can be a mapping `False => None`.

PR #401, opened 2018-06-08
Use template=no,- pattern consequently

gronke:
addresses #377

- `template=-` equals `template=no` in any place of the application


gronke:
> Do we have a place where we could just put `"template=no,-" as a constant?

I suspect that to decrease the readability because other filters are explicitly written as string. We should also keep in mind that this is exclusively implemented in the CLI, not the library.

PR #411, opened 2018-06-11
exit successfully on ioc get -p when pool found

### LINKED ISSUE #407: ioc get --pool exits with 1 — on success
```
/u/l/s/libiocage (master|✔) # python3.6 /usr/local/src/libiocage -d spam get --pool
/etc/rc.conf was read from /etc/rc.conf
Updated /etc/rc.conf data from /etc/rc.conf
Found active ZFS pool zroot
/etc/rc.conf was read from /etc/rc.conf
Updated /etc/rc.conf data from /etc/rc.conf
Found active ZFS pool zroot
zroot
/u/l/s/libiocage (master|✔) # echo $status
1
/u/l/s/libiocage (master|✔) #
```

igalic:
previously, `ioc get -p` would always exit 1, regardless of success.
This fixes #407

### DIFF
--- a/iocage/cli/get.py
+++ b/iocage/cli/get.py
@@ -63,11 +63,13 @@ def cli(
     host = iocage.lib.Host.Host(logger=logger)
 
     if _pool is True:
+        exit_status = 0
         try:
             print(host.datasets.active_pool.name)
         except Exception:
+            exit_status = 1
             print("No active pool found")
-        exit(1)
+        exit(exit_status)
 
     if jail == "":
         prop = ""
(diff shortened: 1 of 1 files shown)


iocage/cli/get.py
```diff
@@ -63,11 +63,13 @@ def cli(
     host = iocage.lib.Host.Host(logger=logger)
 
     if _pool is True:
+        exit_status = 0
         try:
             print(host.datasets.active_pool.name)
         except Exception:
+            exit_status = 1
```
gronke: we can exit here directly with `exit(1)` and exit with 0 later on so that the exit_status variable is not required.

igalic: fair

PR #412, opened 2018-06-11
consistently return - instead of None

### LINKED ISSUE #408: ioc get all defaults deviate from default set on jail creation
the `defaults` use `None` in several places, where a freshly created jail uses `-`

```sh
% ioc get all defaults | grep None
id:None
jail_zfs_dataset:None
```

igalic:
the baseconfig now too returns "-" instead None
This fixes #408

### DIFF
--- a/iocage/lib/Config/Jail/BaseConfig.py
+++ b/iocage/lib/Config/Jail/BaseConfig.py
@@ -150,7 +150,7 @@ class BaseConfig(dict):
             self.legacy = False
 
     def _get_id(self) -> str:
-        return str(self.data["id"])
+        return iocage.lib.helpers.to_string(self.data["id"])
 
     def _set_id(self, name: str, **kwargs) -> None:  # noqa: T484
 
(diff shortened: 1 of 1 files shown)


iocage/lib/Config/Jail/BaseConfig.py
```diff
@@ -150,7 +150,7 @@ def _set_legacy(  # noqa: T484
             self.legacy = False
 
     def _get_id(self) -> str:
-        return str(self.data["id"])
+        return iocage.lib.helpers.to_string(self.data["id"])
```
igalic: oops

```
./iocage/lib/Config/Jail/BaseConfig.py:153:10: T484 Returning Any from function declared to return "str"
```

iocage/lib/Config/Jail/BaseConfig.py
```diff
@@ -382,7 +382,7 @@ def _set_vnet(  # noqa: T484
 
     def _get_jail_zfs_dataset(self) -> typing.List[str]:
         try:
-            jail_zfs_dataset = str(self.data["jail_zfs_dataset"])
+            jail_zfs_dataset = iocage.lib.helpers.to_string(self.data["jail_zfs_dataset"])
```
igalic: double oops

```
./iocage/lib/Config/Jail/BaseConfig.py:385:80: E501 line too long (90 > 79 characters)
./iocage/lib/Config/Jail/BaseConfig.py:386:14: T484 Returning Any from function declared to return "List[str]"
```

iocage/lib/Config/Jail/BaseConfig.py
gronke: Just wrap `str()` around it - that should satisfy the type check.

igalic:
well, in that case… *closes pr*

PR #417, opened 2018-06-16
Implement BaseConfig.get() method

gronke:
This allows requesting jail config properties or fallback to a default. e.g.

```
import iocage
jail = iocage.Jail("myjail")
jail.config.get("prop", False)
```

I would expect some changes to the way user defined properties work, so that this method might not last very long. Discussion is still ongoing and in the meanwhile this is a proper way to deal with unset user default properties for automation tasks.

### DIFF
--- a/iocage/lib/Config/Jail/BaseConfig.py
+++ b/iocage/lib/Config/Jail/BaseConfig.py
@@ -518,6 +518,10 @@ class BaseConfig(dict):
         *args: typing.Any
     ) -> typing.Any:
         """Return the config value or its given default."""
+        if len(args) > 1:
+            raise TypeError(
+                f"get() takes 1 positional argument but {len(args)} were given"
+            )
         try:
             return self.__getitem__(key)
         except KeyError:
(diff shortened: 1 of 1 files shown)


iocage/lib/Config/Jail/BaseConfig.py
```diff
@@ -512,6 +512,23 @@ def _getitem_user(self, key: str) -> typing.Any:
 
         raise KeyError(f"User defined property not found: {key}")
 
+    def get(
+        self,
+        key: str,
+        *args: typing.Any
+    ) -> typing.Any:
+        """Return the config value or its given default."""
+        if len(args) > 1:
```
igalic:
```python
>>> len(None)
TypeError: object of type 'NoneType' has no len()
```

igalic: yeah, args is an array…

gronke: Due to the `*args` notation the `args` value is always a list but may be empty.

PR #418, opened 2018-06-16
enhance data input of ResourceLimitValue on init

gronke:
- ResourceLimitValue may be initialised with a positional argument (str or int)
- ResourceLimit accepts named properties `amount`, `action` and `per`
- Both inputs (positional and named arguments) may be combined - the named arguments overwrite properties parsed from the positional argument.


igalic:
> Both inputs (positional and named arguments) may be combined

that sounds like potential for conflict…

gronke:
>> Both inputs (positional and named arguments) may be combined

> that sounds like potential for conflict…

This is used to allow parsing from a ResourceLimit string instead of passing `amount`, `action` and `per` to the ResourceLimitValue explicitly. When passing a ResourceLimit string it might still be wanted to override a specific attribute.

PR #421, opened 2018-06-21
DHCP without VNET

gronke:
addresses https://github.com/iocage/iocage/issues/404

- move dhcp and rtsold execution from NetworkInterface to Jail

### DIFF
--- a/iocage/lib/NetworkInterface.py
+++ b/iocage/lib/NetworkInterface.py
@@ -37,8 +37,6 @@ class NetworkInterface:
     """
 
     ifconfig_command = "/sbin/ifconfig"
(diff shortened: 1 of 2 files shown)


iocage/lib/NetworkInterface.py
```diff
@@ -37,8 +37,6 @@ class NetworkInterface:
     """
 
     ifconfig_command = "/sbin/ifconfig"
-    dhclient_command = "/sbin/dhclient"
-    rtsold_command = "/usr/sbin/rtsold"
```
igalic: `rtsold_command` is being used below in line 200.

PR #426, opened 2018-07-07
Feature: Provisioning (ix-iocage-plugin)

### LINKED ISSUE #1: Feature: Plugins
Iocage implements a mechanism to automatically install jails from a template called "plugin". Various plugins are listed in the [freenas plugin registry](https://github.com/freenas/iocage-ix-plugins).

The way of installing packages and applying post-install hooks can be abstracted as discussed in https://github.com/iocage/iocage/issues/163 about Ansible (or other provisioning tools) integration. Running a hook script read from a GitHub source would be sufficient to support the current plugin system.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/48437423-feature-plugins?utm_campaign=plugin&utm_content=tracker%2F70027116&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F70027116&utm_medium=issues&utm_source=github).
</bountysource-plugin>

gronke:
closes #1 

- introduces nested jail config properties
- embeds provisioning methods:
  - `ix` for ix-iocage-plugins
  -  ~~`ansible` as runner for ansible playbooks stored in a remote git repository~~ will be implemented in another PR

### ix-iocage-plugin install stages

When a jail is configured with `provisioning.method=ix`, the following will happen when running the provisioning command with `provisioning.source=jenkins`:

1. Download https://github.com/freenas/iocage-ix-plugins/blob/master/jenkins.json
2. Read `artifacts` from jenkins.json
3. Clone artifacts repo URL to a child dataset of the jail
4. Start jail with artifacts mounted
5. Install list of `pkgs` in jenkinsk.json within the jail
6. Run `post_install.sh` from the artifacts folder if it exists

### CLI Syntax

```sh
ioc create plextemplate \
    provisioning.method=ix \
    provisioning.source=plexmediaserver \
    template=yes

# run provisioner
ioc provision plextemplate

# create jail from the provisioned template
ioc create -t plextemplate plexjail

# configure network and run plexjail
# ...
```

**Update**: Providing a temporary network connecting is no longer required. The ix-iocage-plugin now installs packages via [iocage.lib.Pkg](https://github.com/iocage/libiocage/blob/master/iocage/lib/Pkg.py) which does not require internet connections within a jail.

### DIFF
--- a/iocage/cli/provision.py
+++ b/iocage/cli/provision.py
@@ -1,0 +1,94 @@
+# Copyright (c) 2014-2018, iocage
+# All rights reserved.
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted providing that the following conditions
+# are met:
+# 1. Redistributions of source code must retain the above copyright
+#    notice, this list of conditions and the following disclaimer.
+# 2. Redistributions in binary form must reproduce the above copyright
+#    notice, this list of conditions and the following disclaimer in the
+#    documentation and/or other materials provided with the distribution.
+#
+# THIS SOFTWARE IS PROVIDED BY THE AUTHOR ``AS IS'' AND ANY EXPRESS OR
+# IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
+# WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+# ARE DISCLAIMED.  IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR ANY
+# DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+# DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS
+# OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION)
+# HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT,
+# STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING
+# IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+# POSSIBILITY OF SUCH DAMAGE.
+"""Provision jails from the CLI.."""
+import typing
+import click
+
+import iocage.lib.errors
+import iocage.lib.Jails
+import iocage.lib.Logger
+
+from .shared.click import IocageClickContext
+
+__rootcmd__ = True
+
+
+@click.command(name="start", help="Trigger provisioning of jails.")
+@click.pass_context
+@click.argument("jails", nargs=-1)
+def cli(
+    ctx: IocageClickContext,
+    jails: typing.Tuple[str, ...]
+) -> None:
+    """Run jail provisioner as defined in jail config."""
+    logger = ctx.parent.logger
+    start_args = {
+        "zfs": ctx.parent.zfs,
+        "host": ctx.parent.host,
+        "logger": logger,
+        "print_function": ctx.parent.print_events
+    }
+
+    if not _provision(jails, **start_args):
+        exit(1)
+
+
+def _provision(
+    filters: typing.Tuple[str, ...],
+    zfs: iocage.lib.ZFS.ZFS,
+    host: iocage.lib.Host.HostGenerator,
+    logger: iocage.lib.Logger.Logger,
+    print_function: typing.Callable[
+        [typing.Generator[iocage.lib.events.IocageEvent, None, None]],
+        None
+    ]
+) -> bool:
+
+    jails = iocage.lib.Jails.JailsGenerator(
+        logger=logger,
+        zfs=zfs,
+        host=host,
+        filters=filters
+    )
+
+    changed_jails = []
+    failed_jails = []
+    for jail in jails:
+        try:
+            print_function(jail.provisioner.provision())
+        except iocage.lib.errors.IocageException:
+            failed_jails.append(jail)
+            continue
+
+        changed_jails.append(jail)
+
+    if len(failed_jails) > 0:
+        return False
+
+    if len(changed_jails) == 0:
+        jails_input = " ".join(list(jails))
+        logger.error(f"No jails started your input: {jails_input}")
+        return False
+
+    return True
--- a/iocage/cli/provision.py
+++ b/iocage/cli/provision.py
@@ -30,6 +30,7 @@ import iocage.lib.Jails
 import iocage.lib.Logger
 
 from .shared.click import IocageClickContext
+from .shared.jail import set_properties
 
 __rootcmd__ = True
 
@@ -37,9 +38,16 @@ __rootcmd__ = True
 @click.command(name="start", help="Trigger provisioning of jails.")
 @click.pass_context
 @click.argument("jails", nargs=-1)
+@click.option(
+    "--option", "-o",
+    "temporary_config_override",
+    multiple=True,
+    help="Temporarily override jail config options"
+)
 def cli(
     ctx: IocageClickContext,
-    jails: typing.Tuple[str, ...]
+    jails: typing.Tuple[str, ...],
+    temporary_config_override: typing.Tuple[str, ...]
 ) -> None:
     """Run jail provisioner as defined in jail config."""
     logger = ctx.parent.logger
@@ -50,12 +58,17 @@ def cli(
         "print_function": ctx.parent.print_events
     }
 
-    if not _provision(jails, **start_args):
+    if not _provision(
+        filters=jails,
+        temporary_config_override=temporary_config_override,
+        **start_args
+    ):
         exit(1)
 
 
 def _provision(
     filters: typing.Tuple[str, ...],
+    temporary_config_override: typing.Tuple[str, ...],
     zfs: iocage.lib.ZFS.ZFS,
     host: iocage.lib.Host.HostGenerator,
     logger: iocage.lib.Logger.Logger,
@@ -76,7 +89,14 @@ def _provision(
     failed_jails = []
     for jail in jails:
         try:
-            print_function(jail.provisioner.provision())
+            set_properties(
+                properties=temporary_config_override,
+                target=jail
+            )
+        except iocage.lib.errors.IocageException:
+            exit(1)
+        try:
+            print_function(_execute_provisioner(jail))
         except iocage.lib.errors.IocageException:
             failed_jails.append(jail)
             continue
@@ -92,3 +112,12 @@ def _provision(
         return False
 
     return True
+
+
+def _execute_provisioner(
+    jail: 'iocage.lib.Jail.JailsGenerator'
+) -> typing.Generator['iocage.lib.events.IocageEvent', None, None]:
+    for event in jail.provisioner.provision():
+        if isinstance(event, iocage.lib.events.JailLaunch) and event.done:
+            print(event.stdout)
+        yield event
--- a/iocage/cli/provision.py
+++ b/iocage/cli/provision.py
@@ -95,6 +95,7 @@ def _provision(
             )
         except iocage.lib.errors.IocageException:
             exit(1)
+
         try:
             print_function(_execute_provisioner(jail))
         except iocage.lib.errors.IocageException:
@@ -118,6 +119,7 @@ def _execute_provisioner(
     jail: 'iocage.lib.Jail.JailsGenerator'
 ) -> typing.Generator['iocage.lib.events.IocageEvent', None, None]:
     for event in jail.provisioner.provision():
-        if isinstance(event, iocage.lib.events.JailLaunch) and event.done:
-            print(event.stdout)
         yield event
+        if isinstance(event, iocage.lib.events.JailCommandExecution):
+            if event.done is True:
+                print(event.stdout)
--- a/iocage/lib/Config/Jail/BaseConfig.py
+++ b/iocage/lib/Config/Jail/BaseConfig.py
@@ -114,9 +114,18 @@ class BaseConfig(dict):
                 self["id"] = data[key]
                 break
 
-        for key, value in data.items():
-            if key in ["id", "name", "uuid"]:
+        current_id = self["id"]
+        items = list(data.items())
+        while len(items) > 0:
+            key, value = items.pop()
+
+            if type(value) == dict:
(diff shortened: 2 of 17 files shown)


gronke:
The `ioc provision` option can be set multiple times and allows to temporary override jail configuration properties. When provisioning a template this can be used to leave the configuration untouched while installing packages.

libiocage does not automatically configure the network when provisioning jails or templates because they are potentially untrusted.

iocage/lib/Provisioning/ix.py
```diff
@@ -148,19 +178,52 @@ def provision(
...
-        "ASSUME_ALWAYS_YES=true pkg update",
-        f"ASSUME_ALWAYS_YES=true pkg install -y {pkg_packages}",
-        "./ix-plugin/post_install.sh"
-    ]
-
-    for event in self.jail.fork_exec("\n".join(commands)):
+    for event in self.jail.start():
         yield event
 
+    try:
```
skarekrow: Why do a `try: finally` block when any of the `try`'s above the `finally` statement would end execution. I think you can drop a whole nested level here.

iocage/lib/Provisioning/ix.py
```diff
@@ -148,19 +178,52 @@ def provision(
...
+            )
+            yield packageUpdateEvent.end(stdout=stdout)
+        except iocage.lib.errors.IocageException as e:
+            yield packageUpdateEvent.fail(e)
+            raise e
+
+        yield packageInstallEvent.begin()
+        try:
+            stdout, stderr, returncode = self.jail.exec(
+                ["pkg", "install", "-yield"] + pkg_packages,
```
skarekrow: `pkg install -yield`?

iocage/lib/Provisioning/ix.py
```diff
@@ -170,8 +233,9 @@ def __get_empty_dataset(
 ) -> libzfs.ZFSDataset:
     try:
         dataset = zfs.get_dataset(dataset_name)
+        dataset.umount()
         zfs.delete_dataset_recursive(dataset)
-    except libzfs.ZFSException:
+    except libzfs.ZFSException as e:
```
skarekrow: You aren't using `e`, you can drop that

iocage/lib/Jail.py
```diff
@@ -375,6 +374,7 @@ def provisioner(self) -> 'iocage.lib.Provisioning.prototype.Provisioner':
         except AttributeError:
             pass
 
+        import iocage.lib.Provisioning
```
skarekrow: Why do you remove the top level import? Recursion issues?

iocage/cli/provision.py
```diff
@@ -118,6 +119,7 @@ def _execute_provisioner(
     jail: 'iocage.lib.Jail.JailsGenerator'
 ) -> typing.Generator['iocage.lib.events.IocageEvent', None, None]:
     for event in jail.provisioner.provision():
-        if isinstance(event, iocage.lib.events.JailLaunch) and event.done:
-            print(event.stdout)
         yield event
+        if isinstance(event, iocage.lib.events.JailCommandExecution):
```
skarekrow: This should just be an additional line with `\` as there isn't anything else to be done if the if statement below isn't `True`

iocage/lib/Provisioning/ix.py
gronke: It would, but instead each nested try block would require the same fallback. This top-level try/finally makes sure that the code in finally is executed independently of the success of its internal steps.

iocage/lib/Provisioning/ix.py
gronke: Hahahaha, damn you auto-correct. 😆(Have no other explanation) 
Of course this must be just `-y`.

iocage/lib/Jail.py
gronke: Yes Python does not like circular dependencies when loading them at import time. And In long term I'd like to reduce the memory footprint and startup time by lazy-loading specific features. There is a relatively small penalty for re-importing dependencies, but this would only happen once per jail if the provisioner is used.

iocage/cli/provision.py
gronke: I don't get it. Could you please further explain or leave a code snippet?

gronke:
@william-gr we've talked about ix-iocage-plugin support at BSDCan. Finally here is the implementation to install the plugins using the provision command (or Jail method).

@skarekrow you have pointed out some plugin features that were not implemented in this branch. In my opinion another application layer than libiocage should implement UI specific features such as printing administration URLs. My recommendation here is to use custom jail config properties to store the desired port and applying it (or the plugins default) during jail start. At jail start you have such properties stored in the environment variables.

Can you spot any missing install steps besides the UI part, that seems FreeNAS specific to me.

iocage/lib/Config/Jail/BaseConfig.py
```diff
@@ -550,7 +581,24 @@ def __getitem__(self, key: str) -> typing.Any:
...
         """Delete a setting from the configuration."""
-        del self.data[key]
+        self.__deleteitem(key, self.data)
+
+    def __deleteitem(
+        self,
+        key: str,
+        data: typing.Dict[str, typing.Any]
+    ) -> None:
+        if "." in key:
```
igalic: *mumbles something about nested json*

iocage/lib/Provisioning/ix.py
```diff
@@ -0,0 +1,241 @@
...
+            yield packageInstallEvent.end(stdout=stdout)
+        except iocage.lib.errors.IocageException as e:
+            yield packageInstallEvent.fail(e)
+            raise e
+
+        yield postInstallExecutionEvent.begin()
+        try:
+            if os.path.isfile(f"{plugin_dataset.mountpoint}/post_install.sh"):
+                stdout, stderr, returncode = self.jail.exec(
+                    ["/.ix-plugin/post_install.sh"]
```
igalic: is this file guaranteed to be executable?

iocage/lib/ResourceUpdater.py
```diff
@@ -441,6 +441,7 @@ class HardenedBSD(Updater):
 
     @property
     def _update_command(self) -> typing.List[str]:
+        return ["sleep", "2000"]
```
igalic: why?

gronke: 😮 Good question. Will be removed!

iocage/lib/Provisioning/ix.py
gronke: It was created by ourselves. Yes, we can assume that this file is existing and executable.

iocage/lib/Config/Jail/BaseConfig.py
gronke: This is about CLI interaction. The internal storage occurs as nested JSON - just the property may be accessed via dot notation.

iocage/cli/provision.py
```diff
@@ -0,0 +1,125 @@
...
+        except iocage.lib.errors.IocageException:
+            exit(1)
+
+        try:
+            print_function(_execute_provisioner(jail))
+        except iocage.lib.errors.IocageException:
+            failed_jails.append(jail)
+            continue
+
+        changed_jails.append(jail)
```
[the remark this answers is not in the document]
gronke: No, at the current time plugins do not signal whether they applied changes or not. We only know that the provisioner ran and did not error.

gronke:
@igalic thank you for the review. The `sleep 2000` finding was a great catch! Your feedback is addressed although it was changed by rebasing the previous commits.

@skarekrow I've tested with several ix-plugins and found the jails starting up properly with the services installed and enabled. Because I personally do not use those plugins to provision my jails it would be great to get feedback from iX side whether the install steps shown here are sufficient.

With the pkg feature #435 we will be able to provision plugins entirely without network access from within the jail. Would be great to know if there are any ix-plugins (planned) that require Internet access during their provisioning.

gronke:
Since rebasing on #435 this feature does no longer require network in a jail.

iocage/cli/provision.py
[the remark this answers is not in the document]
igalic: puppet returns 1 on success, 0 on idempotent success, and 2 on failure

iocage/lib/Provisioning/ix.py
```diff
@@ -0,0 +1,194 @@
...
+# POSSIBILITY OF SUCH DAMAGE.
+"""iocage provisioner for ix-plugins."""
+import typing
+import os.path
+import json
+import urllib.error
+import urllib.request
+import libzfs
+
+import git
```
igalic: back to depending on git

igalic:
@skarekrow ping

iocage/lib/Provisioning/__init__.py
```diff
@@ -0,0 +1,99 @@
...
+			method,
+			logger=self.jail.logger
+		)
+
+	@property
+	def __available_provisioning_modules(
+		self
+	) -> typing.Dict[str, Prototype]:
+		return dict(
+			ix=iocage.lib.Provisioning.ix
```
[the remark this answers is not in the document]
gronke: Yeah, it is the first one that helps defining an API for other provisioners.

iocage/lib/Provisioning/ix.py
gronke: In theory an ix-iocage-plugin repository may not only contain the post_install.sh file but also assets that it ships with. Using gitpython seems the most convenient way to achieve this.

gronke:
- Are there any environment variables that we want to execute the post_install.sh script with? For instance `provisioning.env.DATABASE_NAME="mydatabase"`.
- Can we name a specific location where the service keeps its status information? The link to the administration backend can be a mix of plugin configuration and jail IP address. If a jail has multiple interfaces, it probably also makes sense to print URLs to all of them (and the jails hostname). What I want to say is that printing reasonable output requires deep knowledge about the jails state. At the time of creating a template we know nothing about the situation later on in a jail.

skarekrow:
> My recommendation here is to use custom jail config properties to store the desired port and applying it (or the plugins default) during jail start. At jail start you have such properties stored in the environment variables.

No, this is simply reading a JSON manifest and replacing the IP with the jails IP. This can happen upon initial creation from the template, which overcomplicates things as most users will create a plugin once. There's not much need to have N versions of plex on the same box :D Besides this is currently all handled in iocage, and the user expectation is it continues to quack like the same duck for this.

> Would be great to know if there are any ix-plugins (planned) that require Internet access during their provisioning.

We don't have a roadmap for upcoming plugins besides the parity of the old system to now. The whole point of the system is user plugins. Doing so may require them to have networking, we have to remain open with the design philosophy to enable that.

> Are there any environment variables that we want to execute the post_install.sh script with? For instance provisioning.env.DATABASE_NAME="mydatabase"

No, besides `IOCAGE_IP` (and maybe a couple others I forget :D), they handle all of that in post_install.sh. Perhaps down the line that makes sense, but we haven't run into any situation where that's been necessary

gronke:
> No, this is simply reading a JSON manifest and replacing the IP with the jails IP. This can happen upon initial creation from the template, which overcomplicates things as most users will create a plugin once. There's not much need to have N versions of plex on the same box :D Besides this is currently all handled in iocage, and the user expectation is it continues to quack like the same duck for this.

Iocage does not know which IP is the primary one or the one used for the service. But it is not a big deal to print the info from the JSON after finishing provisioning.

> We don't have a roadmap for upcoming plugins besides the parity of the old system to now. The whole point of the system is user plugins. Doing so may require them to have networking, we have to remain open with the design philosophy to enable that.

You will see the installation fail in case network is missing. We can add steps to setup temporary network to the features documentation.

> No, besides IOCAGE_IP (and maybe a couple others I forget :D), they handle all of that in post_install.sh. Perhaps down the line that makes sense, but we haven't run into any situation where that's been necessary

What is the `IOCAGE_IP`? :neckbeard: 
- Is it the first one that was configured?
- IPv4 or IPv6?
- Couldn't it just be the jails hostname?
- What do we do with DHCP and rtadv?
- The jail knows it's network configuration. Why telling it its own address?

skarekrow:
> Iocage does not know which IP is the primary one or the one used for the service. But it is not a big deal to print the info from the JSON after finishing provisioning.

Sure it does :) The first entry in `ip4_addr`

> You will see the installation fail in case network is missing. We can add steps to setup temporary network to the features documentation.

Yeah I don't see a huge reason to worry about this _yet_, if it's a requirement, it can always be added later

> What is the IOCAGE_IP?

See first answer

> Is it the first one that was configured?

Yep!

> IPv4 or IPv6?

IPv4 right now, as most services listen on that by default

> Couldn't it just be the jails hostname?

This presupposes working DNS on the customers network, IP addresses do not

> What do we do with DHCP and rtadv?

We show that address instead of the `ip4_addr`

> The jail knows it's network configuration. Why telling it its own address?

It's not telling itself, it's telling the user the IP and port of the plugin. For instance `192.168.3.40:32400` for plex. The user may not be aware of the plugins port in many cases. In addition if we use DHCP/rtadv, they won't know the IP address the jail just received, unless they have reservations.

All in all this is for user info, not the jail or environment. It's just to make the human who typed the commands life a *bit* easier :D

gronke:
Selecting the first configured IP address does not seem unreasonable, but kind of random.

The real issue is that there is no right moment to print the message (speaking of `ioc`). When provisioning a template entirely without network access using the `ioc provision` command does not reveal useful information. Later on, when cloning a jail from that template, iocage does not know about the plugin being provisioned. Even though it could know from the configuration that provisioning is set up, it does not mean that the jail was actually provisioned.

A little background on the way IocageEvents are handled. The CLI wraps each command in the [print_event](https://github.com/iocage/libiocage/blob/master/iocage/cli/__init__.py#L77) function while assuming that the entire library does not create print output. The right way to communicate such information needs to be attached to an IocageEvent. Combining this with the previous paragraph means that a provisioned jail needs different output on start (the only time we know its real IP address) but we cannot differentiate a provisioned jail from a non-provisioned jail.

The only way I see to implement this cleanly is to add a jail motd mechanism that allows a starting jail to write to a specific file that is printed after starting it. If we decide that we need this feature I suggest to implement it in a different branch.

All this is no issue for using it in contexts like FreeNAS. If the jail runner knows that it was recently provisioned reading the message from the plugins json file is no big deal, but not done in libiocage.

gronke:
This basic implementation of ix-iocage-plugins also lays the groundwork for libiocage provisioning.

As discussed offline with @skarekrow there are changes to the ix-iocage-plugin standard planned aiming for:
- Compatibility with different OS versions
- Idempotence

Now is the right time to start an open discussion around this topic, so that we end up with a solution that shares interfaces with Ansible and Puppet provisioning.

With support for provisioning of current ix-iocage-plugins and the base implementation for generalised provisioning in this branch is ready to get merged at this point.

PR #433, opened 2018-07-15
Fix HardenedBSD release updates

gronke:
- check before downloading a new release if the patchlevel snapshot already exists
- take release snapshots after applying them, not after fetching
- correct/enhance hbsd-update arguments


gronke:
Labeled this as security issue because existing jails might have an outdated userland without giving proper notice.

PR #435, opened 2018-07-19
Feature: pkg

gronke:
Jail package management:
- Globally mirror used packages
- Install packages to jails from offline repository / cache
- Mirror packages for every OS major release version
- Ignore install notes from packages

Quality Assurance:
- [x] Tested on FreeBSD 11.2
- [x] Tested on HardenedBSD
- [x] Verify `graylog` service (our test candidate for services that require a TTY during startup) can be installed and starts up
- [x] Processes in a Jail console still receive control codes (e.g. Ctrl + C)

### DIFF
--- a/iocage/cli/pkg.py
+++ b/iocage/cli/pkg.py
@@ -22,7 +22,7 @@
 # STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING
 # IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
 # POSSIBILITY OF SUCH DAMAGE.
-"""Console subcommand for the CLI."""
+"""Jail package management subcommand for the CLI."""
 import click
 
 import iocage.lib.Jail
@@ -38,7 +38,7 @@ __rootcmd__ = True
 @click.argument("jail")
 @click.argument("packages", nargs=-1)
 def cli(ctx, jail, packages):
-    """Run jexec to login into the specified jail."""
+    """Manage packages within jails using an offline mirror."""
     logger = ctx.parent.logger
 
     try:
--- a/iocage/cli/pkg.py
+++ b/iocage/cli/pkg.py
@@ -54,6 +54,7 @@ def cli(ctx, jail, packages):
     try:
         pkg = iocage.lib.Pkg.Pkg(
             logger=logger,
+            zfs=ctx.parent.zfs,
             host=ctx.parent.host
         )
         events = pkg.fetch_and_install(
--- a/iocage/cli/pkg.py
+++ b/iocage/cli/pkg.py
@@ -1,0 +1,66 @@
+# Copyright (c) 2014-2018, iocage
+# Copyright (c) 2017-2018, Stefan Grönke
+# All rights reserved.
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted providing that the following conditions
+# are met:
+# 1. Redistributions of source code must retain the above copyright
+#    notice, this list of conditions and the following disclaimer.
+# 2. Redistributions in binary form must reproduce the above copyright
+#    notice, this list of conditions and the following disclaimer in the
+#    documentation and/or other materials provided with the distribution.
+#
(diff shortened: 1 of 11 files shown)


igalic:
> Install packages to jails from offline repository / cache

AAaaaaaaaaaaaaaaaAAAAAAAAAAAAaaaaaaa

iocage/cli/pkg.py
```diff
@@ -0,0 +1,61 @@
...
+
+import iocage.lib.Jail
+import iocage.lib.Logger
+import iocage.lib.errors
+
+__rootcmd__ = True
+
+
+@click.command(name="pkg", help="Manage packages in a jail.")
+@click.pass_context
```
igalic: is this new? did i miss this??

iocage/cli/pkg.py
```diff
@@ -0,0 +1,61 @@
...
+@click.argument("jail")
+@click.argument("packages", nargs=-1)
+def cli(ctx, jail, packages):
+    """Run jexec to login into the specified jail."""
```
igalic: this comment is wrong

iocage/cli/pkg.py
```diff
@@ -0,0 +1,61 @@
...
+# WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+# ARE DISCLAIMED.  IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR ANY
+# DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+# DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS
+# OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION)
+# HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT,
+# STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING
+# IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+# POSSIBILITY OF SUCH DAMAGE.
+"""Console subcommand for the CLI."""
```
igalic: this command is wrong (and i suppose it's also wrong in exec.py ;)

iocage/lib/Pkg.py
```diff
@@ -0,0 +1,208 @@
...
+            yield packageConfigurationEvent.fail(e)
+            raise e
+        yield packageConfigurationEvent.end()
+
+        yield packageInstallEvent.begin()
+        try:
+            self._temporary_jail.fork_exec(" ".join([
+                "pkg",
+                "install",
+                "-r", "libiocage"
```
igalic: i had to look this option up… so let's maybe use --long-form in our programs.

iocage/lib/Pkg.py
```diff
@@ -0,0 +1,208 @@
...
+            logger=self.resource.logger
+        )
+
+        yield packageInstallEvent.begin()
+        try:
+            iocage.lib.helpers.exec([
+                "pkg",
+                "fetch",
+                "-d",
+                "-o", self.dataset.mountpoint,
```
igalic: please use the `--long-options` for documentation purposes.

iocage/lib/Pkg.py
```diff
@@ -0,0 +1,208 @@
...
+        jail_directory = "/usr/local/etc/pkg/repos/"
+        host_directory = f"{jail.root_path}/{jail_directory}"
+
+        iocage.lib.helpers.makedirs_safe(
+            host_directory,
+            logger=self.logger
+        )
+
+        repo_config_data = {
+            "libiocage": {
```
igalic: starting to think this should be a constant of this class…

gronke:
```sh
ioc create pkgtest
ioc pkg pkgtest vim-console git
```

iocage/cli/pkg.py
gronke: Since a while instances of key classes are created in the CLI tool globally. This way we can use them in any CLI command.

PR #437, opened 2018-07-20
correct key of the temporary jail property exec_start

gronke:
Fixes an issue where jail services were started before the update process. The jail config property `exec.start` does not exist and was corrected to `exec_start`. Services are no longer started during updates of releases and jails.

### DIFF
--- a/iocage/lib/ResourceUpdater.py
+++ b/iocage/lib/ResourceUpdater.py
@@ -122,7 +122,7 @@ class Updater:
                     "basejail": False,
                     "allow_mount_nullfs": "1",
                     "release": self.release.name,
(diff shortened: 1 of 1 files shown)


iocage/lib/ResourceUpdater.py
```diff
@@ -122,7 +122,7 @@ def temporary_jail(self) -> 'iocage.lib.Jail.JailGenerator':
                     "basejail": False,
                     "allow_mount_nullfs": "1",
                     "release": self.release.name,
-                    "exec.start": None,
+                    "exec_start": None,
```
igalic: this should be fixed … generally.

PR #438, opened 2018-07-23
IP address validation

### LINKED ISSUE #371: consider validating ip4 / ip6 addresses on set

gronke:
fixes #371 

- Uses Python ipaddress module

### DIFF
--- a/iocage/lib/Config/Jail/Properties/Addresses.py
+++ b/iocage/lib/Config/Jail/Properties/Addresses.py
@@ -23,6 +23,8 @@
 # POSSIBILITY OF SUCH DAMAGE.
 """Jail config address property."""
 import typing
+import ipaddress
+
 import iocage.lib.errors
 import iocage.lib.helpers
 
@@ -30,11 +32,32 @@ import iocage.lib.helpers
 import iocage.lib.Config.Jail
 import iocage.lib.Logger
 
+IPv4AddressInput = typing.Union[
+    str,
+    ipaddress.IPv4Interface
+]
+
+IPv6AddressInput = typing.Union[
+    str,
+    ipaddress.IPv6Interface
+]
+
+IPAddressInput = typing.Union[
+    str,
+    ipaddress.IPv4Interface,
+    ipaddress.IPv6Interface
+]
+
+IPInterfaceList = typing.Union[
+    typing.List[ipaddress.IPv4Interface],
+    typing.List[ipaddress.IPv6Interface]
+]
+
 
 class AddressSet(set):
     """Set of IP addresses."""
 
-    config: 'iocage.lib.Config.Jail.JailConfig.JailConfig'
+    config: typing.Optional['iocage.lib.Config.Jail.JailConfig.JailConfig']
 
     def __init__(
         self,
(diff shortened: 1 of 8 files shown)


iocage/lib/Config/Jail/Properties/Addresses.py
```diff
@@ -43,27 +53,35 @@ def __init__(
...
+
+    def __parse_address(
+        self,
+        address: IPAddressInput
+    ) -> typing.Union[ipaddress.IPv4Interface, ipaddress.IPv6Interface]:
+        if isinstance(address, str) is True:
+            address = address.lower()
+            if address in ["accept_rtadv", "dhcp"]:
+                return address
+        return ipaddress.ip_interface(address)
```
igalic: i'm confused about this function name… but i looked it up and it makes perfect sense

but what we now don't actually handle is the [ValueError, which ip_interface raises](https://docs.python.org/3/library/ipaddress.html#ipaddress.ip_interface) if it encounters an error.

iocage/lib/Config/Jail/Properties/Addresses.py
```diff
@@ -154,6 +169,37 @@ def add(
...
+    ) -> None:
+
+        if isinstance(addresses, list) is False:
+            self.add(nic=nic, addresses=[addresses], notify=notify)
+            return
+
+        _addresses: typing.List[IPAddressInput] = []
+        try:
+            _addresses = [self.ADDRESS_CLASS(address) for address in addresses]
+        except ipaddress.AddressValueError as e:
```
igalic: i think this should simply be ValueError, at least according to the documentation

igalic:
looking at Travis Errors now, i only now realize how very work-in-progress this whole thing is ;_)

gronke:
When I label a PR as "Work in Progress" you're looking at bare drafts. We can discuss the direction, but please do not expect them to be functional before the label disappears.

In this case I had to pause my work because a couple of bugs needed attention first. 🐞

iocage/lib/Config/Jail/Properties/Addresses.py
gronke: There are three kinds of address-types supported - Host, Net and Interface.

- Host: may not include a netmask
- Net: may only refer to the network address + netmask
- Interface: stores full client address `<IP>/<mask>`

iocage/lib/Config/Jail/Properties/Addresses.py
gronke:
```
>>> import ipaddress
>>> ipaddress.IPv6Interface("192.168.23.1/24")
  File "/usr/local/lib/python3.6/ipaddress.py", line 2077, in __init__
    IPv6Address.__init__(self, addr[0])
  File "/usr/local/lib/python3.6/ipaddress.py", line 1917, in __init__
    self._ip = self._ip_int_from_string(addr_str)
  File "/usr/local/lib/python3.6/ipaddress.py", line 1654, in _ip_int_from_string
    raise AddressValueError(msg)
ipaddress.AddressValueError: At least 3 parts expected in '192.168.23.1'
```

gronke:
@igalic the branch is updated and the WIP flag removed. 🤓

PR #440, opened 2018-07-23
defaults for host_hostname and host_domainname

gronke:
The jail properties `host_hostname` and `host_domainname` were reported to be unknown. This prevented them from being set or changed and reported warnings when listing jails with those properties defined.

```
The config property 'host_hostname' of jail 'my-jail' is unknown.
```


iocage/lib/Config/Jail/Defaults.py
```diff
@@ -86,7 +86,8 @@ class JailConfigDefaults(iocage.lib.Config.Jail.BaseConfig.BaseConfig):
         "ip6_saddrsel": 1,
         "ip6_addr": None,
         "resolver": "/etc/resolv.conf",
-        "host_domainname": "none",
+        "host_hostname": None,
+        "host_domainname": None,
```
igalic: i'm wondering if we shouldn't set 'local' **here**

gronke: Fine by me. I'd love to see a PR. @skarekrow what's your opinion?

PR #446, opened 2018-07-28
Fix static interface routes for defaultrouter6

gronke:
Using the static interface route syntax defaultrouter6="<gateway>@<interface>" did not result in properly configured routes. The `-6` flag was missing on the route command.


gronke:
Implemented in #438.

PR #456, opened 2018-08-02
escape network interface names and properties

gronke:
Network interface names may contain arbitrary UTF-8 characters. Quotation with shlex.quote mitigates command injection via network interface names or interface properties.

### Proof of Concept
```
ioc set vnet=on interfaces=";cat /etc/passwd" ... myjail
```


gronke:
Correct. Length validation does not have an impact on command injection issues. Soon you will find a PR addressing length validation and shortening interface suffixes used for the [Secure VNET](https://github.com/iocage/libiocage/pull/223) feature.

PR #460, opened 2018-08-05
improve release name parsing

### LINKED ISSUE #459: Rolling update
This is more of a question what would be best way to use ioc with combination of blue/green jail deployment. Creating jail with the same name but different tag will not work so I was trying with different naming schema and it turns out that anything other than [a-z0-9] and dash is not supported. I was able to use . within the jail name, but then I was unable to start the jail. Next thing come up to my mind was using multi pool scenario but then I got stuck with `ioc fetch -r blue/11.2-RELEASE` with the error that release *blue/11.2-RELEASE* is not available (<--- this is probably a bug?).

Anyone else using something similar. 

FYI at the moment I'm using a proprietary wrapper that handle the naming -> UUID translation and use uuid as jail name.

gronke:
fixes #459 

Release names may optionally contain a source dataset name (Multi-Pool) and a patch level.

`[<source_dataset_name>/]<release_name>[-p<patch-level>]`

For example:

```
11.1-RELEASE-p9
storage/11.2-RELEASE-p0
storage/11.2-RELEASE
```


urosgruber:
Thanks for quick turnover. Just checked with the branch and seems to be ok.

iocage/lib/Release.py
```diff
@@ -280,13 +280,16 @@ def name(self) -> str:
...
     def name(self, value: str) -> None:
         """Set the releases identifier (optionally including patchlevel)."""
-        patchlevel_match = re.match(r"^(?:.*/)?(.*)-p([0-9]+)$", value)
-        if patchlevel_match is None:
-            self._name = value
-            self.patchlevel = None
-        else:
-            self._name = patchlevel_match[1]
-            self.patchlevel = int(patchlevel_match[2])
+        match = re.match(r"^(?:(.*)/)?(.*)(?:-p([0-9]+))?$", value)
```
igalic: would you mind explaining this regex?
i don't understand the second bit `(.*)` bit

PR #465, opened 2018-08-08
Enhance jail config template setter

### LINKED ISSUE #377: inconsistency between template=no and template=no,- usage

gronke:
closes #377

- Forces template to be a boolean value
- Resolve flake8 T484 exclusions in BaseConfig by dynamically calling setter functions with `skip_on_error` parameter

### DIFF
--- a/iocage/lib/Config/Jail/BaseConfig.py
+++ b/iocage/lib/Config/Jail/BaseConfig.py
@@ -623,6 +623,7 @@ class BaseConfig(dict):
                 return
 
             self._set_data(key, parsed_value)
+            error = None
         except ValueError as err:
             error = iocage.lib.errors.InvalidJailConfigValue(
                 reason=str(err),
@@ -630,8 +631,9 @@ class BaseConfig(dict):
                 logger=self.logger,
                 level=("warn" if (skip_on_error is True) else "error")
             )
-            if skip_on_error is False:
-                raise error
+
+        if (error is not None) and (skip_on_error is False):
+            raise error
 
     def _set_data(
         self,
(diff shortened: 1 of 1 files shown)


iocage/lib/Config/Jail/BaseConfig.py
```diff
@@ -24,6 +24,7 @@
 # POSSIBILITY OF SUCH DAMAGE.
 """The common base of jail configurations."""
 import typing
+import inspect
```
igalic: i'm very skeptic of inspect. it seems more like a debug module…

iocage/lib/Config/Jail/BaseConfig.py
```diff
@@ -616,22 +598,31 @@ def __setitem__(  # noqa: T400
             setter_method_name = f"_set_{key}"
             if setter_method_name in object.__dir__(self):
                 setter_method = self.__getattribute__(setter_method_name)
+                setter_kwargs = dict()
+                _spec = inspect.getfullargspec(setter_method)
+                if ("skip_on_error" in _spec) or (_spec.varkw is not None):
+                    # only pass skip_on_error when method reads it
+                    setter_kwargs["skip_on_error"] = skip_on_error
+                    # do not skip in this method when the setter takes care
+                    skip_on_error = False
```
igalic: this whole section seems a bit over-engeineered
when we can simply demand all setters have skip on error with default of `False`

gronke: You might be right with this. Because no setter function consumes skip_on_error at the moment, I will remove this entirely.

PR #466, opened 2018-08-08
Feature: fstab Hotplugging

### LINKED ISSUE #346: ioc fstab rm does not unmount
if a jail is running and i issue:

```
ioc fstab rm /root foo
```

then shutdown or or reboot the jail
that bind-mount will not be removed until i either do it manually, or reboot the host.

`ioc fstab show foo` will show it's gone, of course, but since `ioc fstab rm` did not umount it, it's still there.

gronke:
- [x] Mounts added entries when a jail is running
- [x] Unmount a destination when a jail is running before removing the line from fstab
- fixes #346 by allowing changes to a running jail
- removed `**kwargs` from setter function headers. Initially these existed because skip_on_error was optionally read by the setters. Since there was no single occasion in the codebase where this it was used, the optional setter argument was removed entirely.

### DIFF
--- a/iocage/lib/Config/Jail/File/Fstab.py
+++ b/iocage/lib/Config/Jail/File/Fstab.py
@@ -30,6 +30,7 @@ import random
 import re
 
 import iocage.lib.helpers
+import iocage.lib.Types
 import iocage.lib.Config.Jail.File
 
 
@@ -248,8 +249,8 @@ class Fstab(
                 continue
 
             new_line = FstabLine({
-                "source": fragments[0],
-                "destination": destination,
+                "source": iocage.lib.Types.AbsolutePath(fragments[0]),
+                "destination": iocage.lib.Types.AbsolutePath(destination),
                 "type": fragments[2],
                 "options": fragments[3],
                 "dump": fragments[4],
@@ -481,6 +482,23 @@ class Fstab(
 
     def __delitem__(self, index: int) -> None:  # noqa: T484
         """Delete an FstabLine at the given index."""
+        deletion_target_line = self.__getitem__(index)
+        jail_name = self.jail.humanreadable_name
+        source = deletion_target_line["source"]
+        destination = deletion_target_line["destination"]
+        if self.jail.running is True:
+            self.logger.verbose(
+                f"Unmounting {destination} from running jail {jail_name}"
+            )
+            iocage.lib.helpers.umount(
+                destination,
+                force=True,
+                logger=self.logger
+            )
+        self.logger.verbose(
+            f"Deleting fstab entry from jail {jail_name}: "
+            f"{source} -> {destination}"
+        )
         real_index = self._get_real_index(index)
         self._lines.__delitem__(real_index)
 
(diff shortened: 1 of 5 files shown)


iocage/lib/Jail.py
```diff
@@ -694,18 +694,13 @@ def _run_hook(self, hook_name: str) -> typing.Optional[
...
-            realpath = os.path.realpath(_dir)
-            if realpath.startswith(self.dataset.mountpoint) is False:
-                raise iocage.lib.errors.SecurityViolationConfigJailEscape(
-                    file=realpath
-                )
-            if os.path.isdir(realpath) is False:
-                os.makedirs(realpath, 0o755)
+        realpath = os.path.realpath(self.launch_script_dir)
+        if realpath.startswith(self.dataset.mountpoint) is False:
+            raise iocage.lib.errors.SecurityViolationConfigJailEscape(
```
igalic: this is pretty cool! (i hadn't noticed it yet)

iocage/lib/Jail.py
```diff
@@ -1565,9 +1553,11 @@ def _launch_single_command_jail(
             "command=/usr/bin/true"
         ]
 
+        _identifier = str(shlex.quote(self.identifier))
+        _jls_command = f"/usr/sbin/jls -j {_identifier} jid"
```
igalic: what does the `jid` at the end of this command mean?

iocage/lib/Config/Jail/File/Fstab.py
```diff
@@ -376,10 +383,13 @@ def add_line(
         if type(line) == FstabLine:
             # destination is always relative to the jail resource
             if line["destination"].startswith(self.jail.root_path) is False:
-                line["destination"] = "/".join([
+
```
igalic: what's with the empty line here?

iocage/lib/Config/Jail/File/Fstab.py
```diff
@@ -343,7 +349,8 @@ def add_line(
         ],
         skip_existing: bool=False,
         replace: bool=False,
-        auto_create_destination: bool=False
+        auto_create_destination: bool=False,
```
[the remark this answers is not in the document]
gronke: I think so because you can either make sure yourself that the directory exists or you may set this neat flag. Consider this default to be interesting for library users but not so much in the CLI context.

iocage/lib/Jail.py
igalic: ah, i was just looking explicitly for `jid`

iocage/lib/Config/Jail/File/Fstab.py
gronke: Uhm.. It's.. It's waiting for removal.

iocage/lib/Config/Jail/File/Fstab.py
```diff
@@ -248,16 +254,16 @@ def parse_lines(
                 continue
 
             new_line = FstabLine({
-                "source": fragments[0],
-                "destination": destination,
+                "source": iocage.lib.Types.AbsolutePath(fragments[0]),
+                "destination": iocage.lib.Types.AbsolutePath(destination),
```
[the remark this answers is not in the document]
gronke: Nothing to worry about 😄
The type can be enhanced and applied more consequently.

PR #467, opened 2018-08-08
Purge *args and **kwargs

### LINKED ISSUE #282: Improve *args and **kwargs typing
Whenever `*args` or `**kwargs` are used as function arguments we've added a `# noqa: T484` comment, so that flake8 does not enforce it to be typed.

Instead we should replace all such occasions with explicitly named parameters or else use list/dict types.

gronke:
closes #282

In more occasions than required libiocage uses *args and **kwargs arguments. Disabling flake8 complaints by setting `# noqa: T484` led to ignorance of other typing issues.

This chore removes `*args` and `**kwargs` from various places.

### DIFF
--- a/iocage/lib/events.py
+++ b/iocage/lib/events.py
@@ -39,6 +39,8 @@ EVENT_STATUS = (
     "failed"
 )
 
+EventUpdateData = typing.Dict[str, typing.Any]
+
 
 class Scope(list):
     """An independent event history scope."""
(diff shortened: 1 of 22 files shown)


iocage/lib/events.py
```diff
@@ -39,6 +39,8 @@
     "failed"
 )
 
+EventUpdateData = typing.Dict[str, typing.Any]
```
igalic: where is this used?

gronke: Is now removed. Thanks for spotting it!

PR #473, opened 2018-08-13
Rename namespace from iocage.lib to iocage

gronke:
- Preparing libiocage to split CLI and Library
- Selected Classes get exposed by hooking the module
    - iocage.Host -> iocage.Host.Host
    - iocage.Distribution -> iocage.Distribution.Distribution
    - iocage.Jails -> iocage.Jails.Jails
    - iocage.Jail -> iocage.Jail.Jail
    - iocage.Releases -> iocage.Releases.Releases
    - iocage.Release -> iocage.Release.Release
- lazy loading of top-level modules
- `iocage` (the python library) and `ioc` (the CLI tool) are separate python modules


iocage/Config/Jail/Properties/Resolver.py
```diff
@@ -28,29 +28,29 @@
...
     none_matches: typing.List[str] = ["/dev/null", "-", ""]
     _entries: typing.List[str] = []
 
     def __init__(
         self,
-        config: 'iocage.lib.Config.Jail.JailConfig.JailConfig',
+        config: 'iocage.Config.Jail.JailConfig.JailConfig',
         property_name: str="resolver",
-        logger: typing.Optional[iocage.lib.Logger.Logger]=None,
+        logger: typing.Optional['iocage.Logger.Logger']=None,
```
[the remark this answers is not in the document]
gronke: The new behavior of MyPy seems to be correct. `iocage.Resolver` depends on `iocage.Jail` which depends on `iocage.Logger` while the logger has a circular dependency on `iocage.Jail`. I cannot answer your question nevertheless.

Makefile
```diff
@@ -28,6 +28,9 @@ install-travis:
 	python3.6 -m pip install flake8-mutable flake8-docstrings flake8-builtins flake8-mypy bandit bandit-high-entropy-string
 uninstall:
 	python3.6 -m pip uninstall -y iocage
+	@if [ -f /usr/local/etc/rc.d/ioc ]; then \
+		rm /usr/local/etc/rc.d/ioc; \
+	fi
```
igalic: how do we ask pip to uninstall our deps?

gronke: We get rid of all dependencies. 😏Done.

That's not (immediately) possible, so that we will have to figure out if and how other projects do it.

setup.py
```diff
@@ -72,22 +83,38 @@ def get_args(cls, dist, header=None):
 setup(
     name='iocage',
     license='BSD',
-    version='0.2.12',
+    version='0.3.0',
```
igalic: shouldn't this be the only place for our version number

ioc/__init__.py
```diff
@@ -183,7 +184,7 @@ def get_command(self, ctx, name):
     help="Globally override the activated iocage dataset(s)"
 )
 @click.command(cls=IOCageCLI)
-@click.version_option(version="0.2.12 09/17/2017", prog_name="ioc")
+@click.version_option(version="0.3.0 2018/08/25", prog_name="ioc")
```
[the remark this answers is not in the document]
gronke: CLI and lib may have different versions. But indeed there should be a mechanism to bump the version in all places. Different PR :)

setup.py
gronke: Could you create an Issue for that? This should be handled in a different PR.

iocage/Config/Jail/File/Fstab.py
```diff
@@ -247,15 +248,28 @@ def parse_lines(
...
                 continue
 
+            source = fragments[0].strip()
             destination = fragments[1].strip()
 
             # skip lines with destinations overlapping self.basejail_lines
             if destination in basejail_destinations:
                 continue
 
+            _backup_prefix = "backup:///"
```
igalic: this seems like a new feature / different bug

should this be here?

gronke: In exports the jails root dataset is replaced with `backup:///` in the fstab file. It could either be detected here or a new type `AbsolutePath` would be required. Replacing it here seemed to be less difficult, although now the logic is spread in two different places.

PR #476, opened 2018-08-19
forward log level in InvalidJailConfigAddress error

gronke:
Fixes an issue raising an IocageException when attempting to set invalid IP address.

### DIFF
--- a/iocage/lib/errors.py
+++ b/iocage/lib/errors.py
@@ -399,7 +399,8 @@ class InvalidJailConfigAddress(InvalidJailConfigValue):
             property_name=property_name,
             jail=jail,
             reason=reason,
-            level=level
+            level=level,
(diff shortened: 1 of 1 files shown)


iocage/lib/errors.py
```diff
@@ -391,13 +391,16 @@ def __init__(
...
+        level: str="error"
     ) -> None:
         reason = f"expected \"<nic>|<address>\" but got \"{value}\""
         super().__init__(
             property_name=property_name,
             jail=jail,
-            reason=reason
+            reason=reason,
+            level=level,
+            logger=logger
```
igalic: there really should be an easier way to do this

gronke: We just got rid of **kwargs for the sake of strong typing.

PR #490, opened 2018-09-02
Use dedicated pkg.conf for package installation

gronke:
The package repo for the iocage release package mirroring is no longer propagated globally. Instead pkg is prefixed to use alternate directories.

```sh
pkg --config /iocage/pkg/11/pkg.conf -- repo-conf-dir /iocage/pkg/11/repos ...
```

### Summary
- Removed invalid logger argument from Pkg events
- Fix failing repo initialization
- Uses a dedicated `pkg.conf` independent from the host configuration

### DIFF
--- a/iocage/Pkg.py
+++ b/iocage/Pkg.py
@@ -67,7 +67,6 @@ class Pkg:
 
         packageFetchEvent = iocage.events.PackageFetch(
             packages=_packages,
-            logger=self.logger,
             scope=event_scope
         )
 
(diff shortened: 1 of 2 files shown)


iocage/Pkg.py
```diff
@@ -186,10 +200,11 @@ def install(
             pkg_archive_name = self._get_latest_pkg_archive(dataset.mountpoint)
             command = "\n".join([
                 "export ASSUME_ALWAYS_YES=yes",
+                "ls -alR /.iocage-pkg",
```
igalic: what's up with this?

gronke: removed in #492

Issue #491, opened 2018-09-02
Exec have issues if command uses -u 


urosgruber:
I'm not sure if it's just local issue with my configuration but it seems that running `ioc exec foobar ps -aux` will result in error "**su: unknown login x**" 

I tried couple of commands where `-u` is used and always the same problem. Issue is also if I use `ioc exec -f foobar`

gronke:
@urosgruber you can put the command in quotes like

```
ioc exec foobar "ps -aux"
```

otherwise you can signify the end of command options with a double dash `--`

```
ioc exec foobar -- ps -aus
```

urosgruber:
@gronke I tried with double quotes and the error was even more weird "**jexec: execvp: ps -aux: No such file or directory**" But with `--` it does work as expected. Maybe something to put in docs or help?

gronke:
@urosgruber done. `ioc exec --help` will now print appropriate hints.

PR #501, opened 2018-09-06
an `exec` command *cannot* be quoted with ""

igalic:
if we pass "ps -aux" to exec, it will literally try to execute the
binary "ps -aux"


igalic:
c.f.: https://github.com/iocage/libiocage/issues/491#issuecomment-417931815

PR #513, opened 2018-09-20
workaround ZFS unjail issue

gronke:
The ZFS unjail command failed in recent attempts to teardown jails with mounted ZFS datasets. This changes work around that issue by manually unjailing the dataset.


igalic:
very interesting!

Issue #520, opened 2018-09-28
exception when any zpool on a system is unavailable


TypedLambda:
the root pool is zroot an is available. a second pool was imported, but no drive available to the system
(pool not present, but still imported / known to the system)

Activating the present pool is not possible (exception)
this should probably be handled (ignore pool).

ioc, version 0.3.0 2018/08/25
```
# ioc activate zroot
```

```python-traceback
Traceback (most recent call last):
  File "/usr/local/bin/ioc", line 10, in <module>
    sys.dd:exit(cli())
  File "/usr/local/lib/python3.6/site-packages/click/core.py", line 722, in __call__
    return self.main(*args, **kwargs)
  File "/usr/local/lib/python3.6/site-packages/click/core.py", line 697, in main
    rv = self.invoke(ctx)
  File "/usr/local/lib/python3.6/site-packages/click/core.py", line 1066, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/usr/local/lib/python3.6/site-packages/click/core.py", line 895, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/usr/local/lib/python3.6/site-packages/click/core.py", line 535, in invoke
    return callback(*args, **kwargs)
  File "/usr/local/lib/python3.6/site-packages/click/decorators.py", line 17, in new_func
    return f(get_current_context(), *args, **kwargs)
  File "/usr/local/lib/python3.6/site-packages/ioc/activate.py", line 62, in cli
    logger=logger
  File "/usr/local/lib/python3.6/site-packages/iocage/Datasets.py", line 187, in __init__
    self._configure_from_pool_property()
  File "/usr/local/lib/python3.6/site-packages/iocage/Datasets.py", line 205, in _configure_from_pool_property
    active_pool = self._active_pool_or_none
  File "/usr/local/lib/python3.6/site-packages/iocage/Datasets.py", line 300, in _active_pool_or_none
    if self.is_pool_active(pool):
  File "/usr/local/lib/python3.6/site-packages/iocage/Datasets.py", line 389, in is_pool_active
    return self._is_pool_active(pool)
  File "/usr/local/lib/python3.6/site-packages/iocage/Datasets.py", line 396, in _is_pool_active
    self.ZFS_POOL_ACTIVE_PROPERTY
  File "/usr/local/lib/python3.6/site-packages/iocage/Datasets.py", line 405, in _get_pool_property
    if prop in pool.root_dataset.properties:
[...]

PR #521, opened 2018-09-28
ignore ZFS pools that are not ONLINE

gronke:
addresses #520

When a ZFS pools status is not `ONLINE`, the pool will be ignored for further actions from iocage. A message is printed to the verbose log level instead.

### DIFF
--- a/iocage/Datasets.py
+++ b/iocage/Datasets.py
@@ -402,7 +402,7 @@ class Datasets(dict):
         prop: str
     ) -> typing.Optional[str]:
 
-        if pool.status != "ONLINE":
+        if (pool.status != "ONLINE") and (pool.status != "DEGRADED"):
             self.logger.verbose(
                 f"The pool {pool.name} is {pool.status} and will be ignored"
             )
--- a/iocage/Datasets.py
+++ b/iocage/Datasets.py
@@ -402,6 +402,12 @@ class Datasets(dict):
         prop: str
     ) -> typing.Optional[str]:
 
+        if pool.status != "ONLINE":
+            self.logger.verbose(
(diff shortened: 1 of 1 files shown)


igalic:
we could `warn` on `DEGRADED`
but perhaps we should document how to setup monitoring / alerting for these not so good states.

gronke:
> Ignoring DEGRADED pools is a mistake IMO. DEGRADED state can last several days depending on the size of the pool.

That's true!

> we could warn on DEGRADED

I don't think this is up to us to warn about it. With the `DEGRADED` state added to the whitelist we should be fine.

This whole discussion raised some questions:
- Should iocage really refuse to interface with jails that are running, but are stored on a ZFS pool that went down. All important assets could come from a NullFS mount so that the jail could still be properly shutdown?
- The Datasets instance is built on startup time of libiocage. How does an iocage daemon get notified about the change? Is it better to lookup the source dataset each time it is accessed or does py-libzfs update the state of the same object instance automatically?
- How good is libiocage in following a dataset to its new mountpoint? A JailGenerator instance could persist a modification of a jail datasets mountpoint. What happens in such occasions? Is there a way to lock datasets while a jail is running?

iocage/Datasets.py
```diff
@@ -402,6 +402,12 @@ def _get_pool_property(
         prop: str
     ) -> typing.Optional[str]:
 
+        if (pool.status != "ONLINE") and (pool.status != "DEGRADED"):
```
skarekrow: This reads nicer as `pool.status not in ('ONLINE', 'DEGRADED')`

gronke: Will address in a separate PR. Good catch!

PR #536, opened 2018-10-04
Release v0.3.1

gronke:
- bump version to 0.3.1


igalic:
would you still call this 0.x?
how about being daring and go for a 1.0.0?

gronke:
According to the iocage/iocage versioning scheme this shoud be even v2.0. I'd go this step when https://github.com/iocage/iocage/tree/v2 became the main branch.

PR #540, opened 2018-10-05
Improve resource limit input

gronke:
- accept rctl syntax resource limit input without /per suffix

### Before
```
ioc . set maxproc=sigkill=50 <JAIL>
Invalid value for property 'maxproc': not enough values to unpack (expected 2, got 1)
```

### After
```
ioc set maxproc=sigkill=50 <JAIL>
ioc set maxproc=sigkill=50/jail <JAIL>
```


igalic:
i just tested this and it works!

Issue #461, opened 2018-08-06
Differences in jail config properties to python-iocage


gronke:
When interacting with jails created with [python-iocage](https://github.com/iocage/iocage/tree/c15be3d74c2e9618fa5708d7a2924a096fa11402) libiocage reports a couple of unknown properties.

```
The config property 'CONFIG_VERSION' of jail 'iptest' is unknown.
The config property 'hostid_strict_check' of jail 'test-bad' is unknown.
The config property 'host_time' of jail 'test-bad' is unknown.
The config property 'dhcp' of jail 'test-bad' is unknown.
The config property 'depends' of jail 'test-bad' is unknown.
The config property 'comment' of jail 'test-bad' is unknown.
The config property 'bpf' of jail 'test-bad' is unknown.
```

@skarekrow could you explain what they are used for?

skarekrow:
Sure, some are in the man page however :P I'll copy the more verbose answers here

'CONFIG_VERSION` - The config revision of iocage they are on :)

`hostid_strict_check` - If used, the jail will refuse to start if the host's hostid is not identical.

`host_time` - Puts the hosts tz into the jail to reduce post configuration

`dhcp` - Enables...DHCP :P
```
     dhcp=[on | off]
                   This controls starting the jail with the Dynamic Host
                   Configuration Protocol enabled.  To enable dhcp, vnet and
                   bpf must also be enabled.

                   Default: off

                   Source: local
```

`depends` - Will start the specified jails before the current jail, great for a FEMP stack.
```
     depends=“none | foo bar”
                   Require another jail to start before starting this jail.
                   Space delimited.  The option nests, resulting in dependent
                   jails waiting in turn for their dependents, if specified,
                   to start.

                   Default: "none"

                   Source: local
```

`comment` - A comment about the jail, not parsed

`bpf` - Passes the bpf device into the jail and unhides it with devfs rules, required for DHCP in a jail cleanly. But has security implications.
```
## IOCAGE -- Add DHCP to ruleset 4
[devfsrules_jail_dhcp=5]
add include $devfsrules_hide_all
add include $devfsrules_unhide_basic
add include $devfsrules_unhide_login
add path zfs unhide
add path 'bpf*' unhide
```

gronke:
Thank you for starting to mention feature differences. We should still put those differences in our [Version Matrix](https://github.com/iocage/libiocage/wiki/Version-Matrix), but I'm glad to comment here as well:

#### DHCP
[...]

PR #541, opened 2018-10-06
Feature: Depends

gronke:
related to #461

- [x] start dependant jails on start of a jail that has dependencies
- [x] stop jails that have been started as dependants if a following dependant fails to start
- [x] stop started dependants if the target jail fails to start
- [x] Resolve dependant jails dependencies to start everything in reasonable order
- [x] dependant jails can be selected using Filter.Terms
- [x] do not start already started jails in `ioc start --rc` command
- [x] resolve circular dependencies

This example shows how the database jail `mydatabase` is started before `mybackend`:

```
ioc create mydatabase tag=database user.customer=gronke
ioc create mybackend depends="tag=database user.customer=gronke" user.customer=gronke
ioc start mybackend
```

### DIFF
--- a/ioc/start.py
+++ b/ioc/start.py
@@ -116,6 +116,9 @@ def _autostart(
     failed_jails = []
     for jail in jails:
         try:
+            if jail.running is True:
+                logger.log(f"{jail.name} is already running - skipping start")
+                continue
             jail.start()
         except iocage.errors.IocageException:
             failed_jails.append(jail)
@@ -162,6 +165,9 @@ def _normal(
             exit(1)
         try:
             jail.require_jail_not_template()
+            if jail.running is True:
+                logger.log(f"{jail.name} is already running - skipping start")
+                continue
             print_function(jail.start())
         except iocage.errors.IocageException:
             failed_jails.append(jail)
--- a/iocage/Config/Jail/Properties/Depends.py
+++ b/iocage/Config/Jail/Properties/Depends.py
@@ -1,0 +1,102 @@
+# Copyright (c) 2014-2018, iocage
+# Copyright (c) 2017-2018, Stefan Grönke
+# All rights reserved.
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted providing that the following conditions
+# are met:
+# 1. Redistributions of source code must retain the above copyright
+#    notice, this list of conditions and the following disclaimer.
+# 2. Redistributions in binary form must reproduce the above copyright
+#    notice, this list of conditions and the following disclaimer in the
+#    documentation and/or other materials provided with the distribution.
+#
+# THIS SOFTWARE IS PROVIDED BY THE AUTHOR ``AS IS'' AND ANY EXPRESS OR
+# IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
+# WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+# ARE DISCLAIMED.  IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR ANY
+# DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+# DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS
+# OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION)
+# HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT,
+# STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING
+# IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+# POSSIBILITY OF SUCH DAMAGE.
+"""Jail config depends property."""
+import typing
+
+import iocage.helpers
+
+
+class DependsProp(iocage.Filter.Terms):
+    """Special jail config property Depends."""
+
+    config: 'iocage.Config.Jail.JailConfig.JailConfig'
+    property_name: str
+    logger: typing.Optional['iocage.Logger.Logger']
+    delimiter: str = ","
+
+    def __init__(
+        self,
+        config: typing.Optional[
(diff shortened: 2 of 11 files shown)


iocage/Filter.py
```diff
@@ -156,21 +158,7 @@ def _filter_string_has_globs(self, filter_string: str) -> bool:
...
-            n = len(values)
-            if n > 0:
-                index = n - 1
-                values[index] += f",{block[0]}"
-            else:
-                values.append(block[0])
-            if len(block) > 1:
-                values += block[1:]
-        return values
+        return re.split(r'(?<!\\),', user_input)
```
igalic: wat

gronke: Split by non-escaped comma.

igalic: please add that as comment

gronke: Moved it to a global variable - it's name steps in instead of a comment.

iocage/Config/Jail/Properties/Depends.py
```diff
@@ -0,0 +1,102 @@
...
+    property_name: str
+    logger: typing.Optional['iocage.Logger.Logger']
+    delimiter: str = ","
+
+    def __init__(
+        self,
+        config: typing.Optional[
+            'iocage.Config.Jail.JailConfig.JailConfig'
+        ]=None,
+        property_name: str="interfaces",
```
igalic: huh? why is the default `str="interfaces"`?

iocage/Config/Jail/Properties/Depends.py
```diff
@@ -0,0 +1,102 @@
...
+
+            notify (bool): (default=True)
+                Sends an update notification to the jail config instance
+        """
+        iocage.Filter.Terms.add(self, term)
+        if notify is True:
+            self.__notify()
+
+    def __delitem__(self, key: typing.Any) -> None:
+        """Remove a jail NIC."""
```
igalic: is this documentation wrong, or what's going on here?

iocage/errors.py
```diff
@@ -370,10 +370,11 @@ class InvalidJailName(JailConfigError):
 
     def __init__(
         self,
+        name: str,
         logger: typing.Optional['iocage.Logger.Logger']=None
     ) -> None:
         msg = (
-            "Invalid jail name: "
+            f"Invalid jail name '{name}': "
```
igalic: how was this not needed before 😅

igalic:
tested this setup:

```sh-session
root@container-host1 /u/l/s/libiocage# ioc list -o name,boot,depends
+----------+------+---------+
|   NAME   | BOOT | DEPENDS |
+==========+======+=========+
| webproxy | yes  | dhcp    |
+----------+------+---------+
| dhcp     | yes  | -       |
+----------+------+---------+
| webirc   | yes  | dhcp    |
+----------+------+---------+
| quassel  | yes  | dhcp    |
+----------+------+---------+
root@container-host1 /u/l/s/libiocage# 
```

with a reboot.

it's not quite clear to me if it's deterministic.
webproxy was attempted to be started first.

iocage/Jail.py
```diff
@@ -578,11 +591,94 @@ def _stop_failed_jail(
...
+            scope=event_scope
+        )
+        started_jails: typing.List[JailGenerator] = []
+
+        yield jailDependantsStartEvent.begin()
+        _depends = self.config["depends"]
+        if len(_depends) == 0:
+            yield jailDependantsStartEvent.skip("No dependant jails")
+        else:
+            dependant_jails = sorted(
```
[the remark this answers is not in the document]
gronke: This part is in the else clause because the other option, no dependant jails, is much shorter and easier to relate to the condition. Will refactor the code to get rid of the else block.

iocage/Jail.py
```diff
@@ -2062,6 +2161,12 @@ def __dir__(self) -> typing.List[str]:
                 properties.add(prop)
         return list(properties)
 
+    def __eq__(self, other: typing.Any) -> bool:
```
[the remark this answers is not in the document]
gronke: This still needs to be improved. We totally ignore the root dataset the jail is located at. Of course a jail with the same name from different iocage root datasets is not equal.

ioc/start.py
```diff
@@ -162,6 +165,9 @@ def _normal(
             exit(1)
         try:
             jail.require_jail_not_template()
+            if jail.running is True:
+                logger.log(f"{jail.name} is already running - skipping start")
+                continue
             print_function(jail.start())
```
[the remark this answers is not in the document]
gronke: This is a workaround when a jail that another jail depended upon was started already that prevents it from being started again while looping over the selected jails.

iocage/Jail.py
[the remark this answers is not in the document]
igalic: oh, then maybe add a `XXX` comment here

iocage/Jail.py
```diff
@@ -578,11 +591,95 @@ def _stop_failed_jail(
...
+            jail=self,
+            scope=event_scope
+        )
+        started_jails: typing.List[JailGenerator] = []
+
+        yield jailDependantsStartEvent.begin()
+        _depends = self.config["depends"]
+        if len(_depends) == 0:
+            yield jailDependantsStartEvent.skip("No dependant jails")
+            return
```
igalic: our return type doesn't state `Optional`

iocage/Jail.py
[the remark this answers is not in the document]
gronke: Docstring is added.

igalic:
rebooted the same server with these new changes and it came back up happy!

urosgruber:
About to test this today as well.

gronke:
> About to test this today as well.

@urosgruber Very nice, thanks. Will wait for your feedback before merging the changes.

urosgruber:
@gronke looks fine, but didn't focus if anything else broke. But I noticed that rename need to be handled manually. So actual dependency is stored in resolved  state, right? 

But displaying depends info kinda confuses me. For example I have a tag on cluster of servers and each have different name. After using `ioc get depends jail` only value is presented without me knowing this is actually a tag value or whatever other attribute I used during `ioc set depends="attr" jail` command. 

And lastly a question. Do I need to have boot=yes on jails that are specified as depends on other jails?

gronke:
The value of the `depends` JailConfig property can be a FilterTerm. The simplest term is a jail name, but you can also add other JailConfig properties to select what jail the jail should depend on.

```console
ioc create prod-database tags=db,production
ioc create dev-database tags=db,staging
ioc create www boot=yes depends="+database tags=production"
```

PR #543, opened 2018-10-08
List active ZFS pools

gronke:
List all active pools maintained by libiocage with `ioc list --pools` instead of just a single pool using `ioc get --pool` (deprecared).

This changes are backported from https://github.com/iocage/iocage/pull/636

### DIFF
--- a/ioc/list.py
+++ b/ioc/list.py
@@ -239,8 +239,9 @@ def _lookup_resource_values(
 ) -> typing.List[str]:
     is_resorce = isinstance(resource, iocage.Resource.Resource)
     if is_resorce and ("getstring" in resource.__dir__()):
+        _resource = resource  # type: iocage.Resource.Resource
         return list(map(
-            lambda column: str(resource.getstring(column)),
+            lambda column: str(_resource.getstring(column)),
             columns
         ))
     else:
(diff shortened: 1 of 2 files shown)


ioc/list.py
```diff
@@ -1,5 +1,4 @@
 # Copyright (c) 2014-2018, iocage
-# Copyright (c) 2017-2018, Stefan Grönke
```
igalic: heh

ioc/list.py
```diff
@@ -206,7 +231,10 @@ def _print_json(
...
 
 def _lookup_resource_values(
-    resource: 'iocage.Resource.Resource',
+    resource: typing.Union[
+        'iocage.Resource.Resource',
+        typing.Dict[str, str]
+    ],
     columns: typing.List[str]
 ) -> typing.List[str]:
     if "getstring" in resource.__dir__():
```
igalic: flake8 error:
```
./ioc/list.py:242:33: T484 Item "Dict[str, str]" of "Union[Any, Dict[str, str]]" has no attribute "getstring"
```

gronke: Flake8 is a merge requirement. No need to report those manually in reviews :)

PR #546, opened 2018-10-10
fix DHCP client usage

gronke:
This PR was created while pairing with @igalic.


igalic:
why are the tests failing?

Issue #552, opened 2018-10-13
some components are still trying to sendmail or at least attemp mailq cleanup


igalic:
According to the [FreeBSD Handbook on how to disable (replace) Sendmail](https://www.freebsd.org/doc/en_US.ISO8859-1/books/handbook/mail-changingmta.html), we also have to disable these:

```
daily_clean_hoststat_enable="NO"
daily_status_mail_rejects_enable="NO"
daily_status_include_submit_mailq="NO"
daily_submit_queuerun="NO"
```

from what i understand, we can also tell [`cron(8)`](https://www.freebsd.org/cgi/man.cgi?cron) to start with `cron_flags="-m ''"` to not mail.

PR #553, opened 2018-10-13
disable mailq cleanups and cron mailing

igalic:
When disabling sendmail, certain periodic(8) components are still kept
running to do the cleanup.
Furthermore, we also disable cron's mailing.

This is a fix for #552

### DIFF
--- a/iocage/Release.py
+++ b/iocage/Release.py
@@ -215,6 +215,11 @@ class ReleaseGenerator(ReleaseResource):
         "sendmail_submit_enable": False,
         "sendmail_msp_queue_enable": False,
         "sendmail_outbound_enable": False,
+        "daily_clean_hoststat_enable": False,
+        "daily_status_mail_rejects_enable": False,
+        "daily_status_include_submit_mailq": False,
+        "daily_submit_queuerun": False,
+        "cron_flags": "-m ''",
         "syslogd_flags": "-ss"
     }
 


iocage/Release.py
```diff
@@ -215,7 +215,12 @@ class ReleaseGenerator(ReleaseResource):
         "sendmail_submit_enable": False,
         "sendmail_msp_queue_enable": False,
         "sendmail_outbound_enable": False,
-        "syslogd_flags": "-ss"
+        "daily_clean_hoststat_enable": False,
+        "daily_status_mail_rejects_enable": False,
+        "daily_status_include_submit_mailq": False,
+        "daily_submit_queuerun": False,
+        "syslogd_flags": "-ss",
```
gronke: you could place syslogd_flags to the end, so that the line does not change

igalic: that would make sense, esp given that the cron line actually groups *with* this

iocage/Release.py
```diff
@@ -215,6 +215,11 @@ class ReleaseGenerator(ReleaseResource):
         "sendmail_submit_enable": False,
         "sendmail_msp_queue_enable": False,
         "sendmail_outbound_enable": False,
+        "daily_clean_hoststat_enable": False,
+        "daily_status_mail_rejects_enable": False,
+        "daily_status_include_submit_mailq": False,
+        "daily_submit_queuerun": False,
```
igalic: i don't know how i got the idea to put these into `rc.conf`, but they belong into [`periodic.conf`](https://www.freebsd.org/cgi/man.cgi?periodic.conf)

PR #554, opened 2018-10-13
remove duplicate .env import from start script when VNET is enabled

### LINKED ISSUE #549: launch-scripts/start.sh sources /.iocage/.env twice
Here's mine:

```sh
#!/bin/sh
. /.iocage/.env
. /.iocage/.env
/sbin/ifconfig vnet0:$IOCAGE_JID:j link 02:ff:60:db:c9:16 mtu 1500 name vnet0 up > /dev/null
/sbin/dhclient vnet0 > /dev/null
/sbin/ifconfig lo0 localhost
/bin/sh /etc/rc
```

gronke:
closes #549

Sourcing the jails .env file happened twice when VNET was enabled. Because the environment variables became imported globally there was no need to do so specifically for the configuration of VNET devices.

### DIFF
--- a/iocage/Jail.py
+++ b/iocage/Jail.py
@@ -483,7 +483,7 @@ class JailGenerator(JailResource):
         if self.config["vnet"]:
             _created, _start = self._start_vimage_network()
             exec_created += _created
-            exec_start += [f". {self._relative_hook_script_dir}/.env"] + _start
+            exec_start += _start
             exec_start += self._configure_localhost_commands()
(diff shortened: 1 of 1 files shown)


gronke:
Thanks to @igalic for [pointing out the solution](https://github.com/iocage/libiocage/issues/549#issuecomment-429570745) 😸

gronke:
@igalic have you tested and compared the output? asking just for curiosity ^^

igalic:
heck no im in bed

Issue #562, opened 2018-10-19
ioc defaults leak into temporary (upgrade) jail when downloading (upgrading) releases


igalic:
same as https://github.com/iocage/libiocage/issues/558 — but for `ioc fetch`
when setting global defaults that influence `exec_*` they will leak into the temporary upgrade jail when doing ioc fetch

gronke:
I think we should approach this on a general level. For example by telling the jail to ignore the defaults. Right now that could be done by overriding the defaults source file, but maybe a dedicated option to ignore the defaults is better to understand.

igalic:
that, and/or a general API for temporary_jail!

gronke:
> that, and/or a general API for temporary_jail!

What would that API look like?

igalic:
so, let's see, right now we have two:

```python
    @property
    def temporary_jail(self) -> 'iocage.Jail.JailGenerator':
        """Temporary jail instance that will be created to run the update."""
        if hasattr(self, "_temporary_jail") is False:
            temporary_name = self.resource.name.replace(".", "-") + "_u"
            temporary_jail = iocage.Jail.JailGenerator(
                {
                    "name": temporary_name,
                    "basejail": False,
                    "allow_mount_nullfs": "1",
                    "release": self.release.name,
                    "exec_start": None,
                    "exec_prestart": None,
                    "exec_poststart": None,
                    "exec_prestop": None,
                    "exec_poststop": None,
                    "exec_timeout": 600,
                    "securelevel": "0",
                    "allow_chflags": True,
                    "vnet": False,
                    "ip4_addr": None,
                    "ip6_addr": None,
                    "defaultrouter": None,
                    "mount_devfs": True,
                    "mount_fdescfs": False
                },
                new=True,
                logger=self.resource.logger,
                zfs=self.resource.zfs,
                host=self.resource.host,
                dataset=self.resource.dataset
            )
            temporary_jail.config.file = "config_update.json"

            root_path = temporary_jail.root_path
            destination_dir = f"{root_path}{self.local_release_updates_dir}"
            temporary_jail.fstab.file = "fstab_update"
            temporary_jail.fstab.new_line(
                source=self.host_updates_dir,
                destination=destination_dir,
                options="rw"
            )
            if os.path.isdir(destination_dir) is False:
                os.makedirs(destination_dir, 0o755)
            temporary_jail.fstab.save()
            self._temporary_jail = temporary_jail
        return self._temporary_jail
```

```
[...]

PR #563, opened 2018-10-19
override global exec_* defaults when upgating resources

igalic:
workaround for #562 until we have a more general solution


gronke:
Closing this PR in favor of #574 which mitigates the issue on a more general level. Instead of having to specify every possible JailConfig property, only the difference between the hardcoded defaults and the desired temporary jail config properties needs to be set.

PR #567, opened 2018-10-22
Jail Release Upgrade

gronke:
python-iocage features upgrading a jail across releases. This branch improves the update progress and the handling of release patchlevels.


gronke:
Some experiments have revealed that the updates we fetch are based on a single release. That means that the current offline patching of iocage jails cannot be used to upgrade releases.

The most efficient way to upgrade a jail to a new release is in my opinion to follow this steps:

1. create a new jail from the target release
2. copy jail config to the new jail
3. clone additional datasets within the jail dataset
4. copy differential data compared to the old release to the new jail
5. run `etcupdate` in the new jail
6. promote additional datasets
7. delete old jail
8. rename new jail

Step 5 potentially requires user interaction, but if it fails and interaction is required. An additional flag on the upgrade command runs the update commands attached to the current TTY, so that users can respond with their input.

gronke:
Closed in favor of #579 - the feature is not trivial and will be re-visited later. @kmoore134 seems to work on a similar issue. Hoping he'll point us in the right direction once he solved it. 👋

PR #577, opened 2018-11-10
refactor Resource namespace

gronke:
- groups Resource in a module directory


iocage/Resource/__init__.py
```diff
@@ -413,3 +413,16 @@ def destroy(
     def save(self) -> None:
         """Save changes to the default configuration."""
         self._write_config(self.config.user_data)
+
+
+iocage.hook_module(
```
igalic: is this a python thing, or a (lib)iocage thing?

igalic (commented):
why is our hierarchy so deep & complicated to begin with?

urosgruber:
Is this done and can be tested or some work still required?

igalic:
> Is this done and can be tested or some work still required?

@urosgruber unfortunately, @gronke has been burried under real life, paid work, so this hasn't been touched since, and i'm fairly certain it needs some more work ***before*** it can be tested.

Unless you feel particularly adventures 😅

urosgruber:
@igalic thanks for the update. Anyone knows how long will it take for this project to continue with it's pace. I notice it was a bit quiet here lately. I would help here but I'm very rookie around Python so I doubt there is anything I can do except testing.

gronke:
As @igalic already has pointed out I've put focus on paid work, so that I was entirely booked in the past two month. This will continue in December and January and is necessary to secure my independence developing the lib, as done the year before. My goal is to avoid the commercial forces that can quickly drift projects like iocage in a one-sided direction. I am not sleeping, just breathing.

The community around libiocage is small, but all the feedback, reviews and contributions motivate me to continue with the same quality objectives in mind. Speaking up here on GitHub is now more important than ever before - please stay alerted on the https://github.com/iocage/iocage/tree/v2 branch as well, because this is where shit will hit the fan.

iocage/Resource/__init__.py
gronke: A local workaround for ugly Python module names. Makes them callable and act like classes with the same names.

urosgruber:
@gronke really appreciate your hard work and quick turnarounds.

igalic:
i still really like this work!
it'll need a rebase now, however, aaaaand a documentation update

PR #578, opened 2018-11-18
libioc should support iocage's template jails

sonicaj:
Libiocage does not right now support iocage's template jails. This commit introduces the following changes:
1) Make sure that libiocage is able to create jails from iocage's template jails
2) Libiocage is able to list iocage's template jails
3) A bug fix in JailConfig where we try to fetch the rc_conf file when the jail does not exist
4) Bug fixes in special properties ( DefaultRouter, Resolver ) where a bug in libiocage did not anticipate that objects of these special properties could be sent as as set value

### DIFF
--- a/iocage/ListableResource.py
+++ b/iocage/ListableResource.py
@@ -25,6 +25,7 @@
 """iocage Resource module."""
 import typing
 import libzfs
+import itertools
 import abc
 
 import iocage.Filter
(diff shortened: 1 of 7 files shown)


iocage/ListableResource.py
```diff
@@ -81,7 +84,7 @@ def __iter__(
         self
     ) -> typing.Generator['iocage.Resource.Resource', None, None]:
         """Return an iterator over the child datasets."""
-        if self.namespace is None:
```
skarekrow: Perhaps we make this an OR with the below addition?

skarekrow (changes requested):

iocage/ListableResource.py
sonicaj: I am not sure if I followed you correctly. Are you suggesting `self.namespace is None or not self.namespace` ?

skarekrow: Yep! Stefan probably has a specific use for None.

sonicaj: Oh, I was not aware of that. `not self.namespace` should normally cover the general negation scenario but if we have a specific use case, then that needs to be followed through. Let's confirm this ? @gronke could you confirm this please ?

PR #579, opened 2018-12-07
Improve update and export

gronke:
replaces #567

- refactoring of Storage classes
- exclude release basedirs from backup
- diff a jail against its release snapshot
- allow to update basejails to minimize overhead on export


iocage/ResourceUpdater.py
```diff
@@ -431,12 +431,11 @@ def _update_jail(
                     "The update jail is still running. "
                     "Force-stopping it now."
                 )
-                for event in iocage.Jail.JailGenerator.stop(
+                yield from iocage.Jail.JailGenerator.stop(
```
igalic: nice!

PR #580, opened 2018-12-13
test libioc on Cirrus CI

igalic:
Cirrus CI offers freebsd builds, so let's use that for our actual tests

mypy can remain on travis as it is orders of magnitude faster!


Makefile
```diff
@@ -3,32 +3,39 @@ JAIL_NIC?=vtnet0
 JAIL_IP?=172.16.0
 JAIL_NET?=16
 MYPYPATH = $(shell pwd)/.travis/mypy-stubs
+PYTHON ?= python3.6
+
+pyver= ${PYTHON:S/^python//:S/.//:C/\([0-9]+\)/\1/}
+
+.if $(pyver) < 35
+. error "libiocage cannot run with a Python version < 3.5"
```
gronke: please rename to libioc

igalic:
> I don't see the direct relation to a CI runner, so maybe the PR should be renamed before merging.

have i ever made any changes that were straight forward? 😝

.cirrus.yml
```diff
@@ -0,0 +1,26 @@
...
+    folder: ~/.cache/pip
+    fingerprint_script: cat requirements*
+    populate_script: make install-python-requirements-dev
+  zfs_cache:
+    folder: /pools/
+    fingerprint_script: cat libioc/VERSION
+    populate_script:
+      - mkdir -p /pools/
+      - '[ -f "/pools/ioc-test-`uname -r`.img" ] && zpool import -d /pools "ioc-test-`uname -r`"'
+      - '[ ! -f "/pools/ioc-test-`uname -r`.img" ] && truncate -s 10G "/pools/ioc-test-`uname -r`.img" && zpool create -m   "/.iocage-test`uname -r`" "ioc-test-`uname -r`" "/pools/ioc-test-`uname -r`.img"'
```
igalic: why is this a single line, btw?

tests/release_mirror_cache.py
```diff
@@ -0,0 +1,74 @@
...
+import threading
+
+LOCAL_MIRROR_PORT=8234
+
+uname = os.uname()
+
+
+class CacheHandler(SimpleHTTPRequestHandler):
+
+    basedir = "/releases"
```
igalic: a better default is `~/releases`

tests/conftest.py
```diff
@@ -165,6 +179,17 @@ def release(
...
+        target_release = "12.0-RELEASE"
+
+    release = libioc.Release.Release(
+        name=target_release, host=host, logger=logger, zfs=zfs
     )
+    return release
+
+
+import release_mirror_cache
+release_mirror_cache.run_thread(8081)
```
igalic: how do we opt out of this?

tests/conftest.py
```diff
@@ -135,27 +128,48 @@ def root_dataset(
...
 
-    if force_clean:
-        helper_functions.unmount_and_destroy_dataset_recursive(dataset)
+
+class MockedDistribution(libioc.Distribution.Distribution):
+
+    @property
+    def mirror_url(self) -> str:
+        print("MIRROR REQUESTED")
+        return "http://127.0.0.1:8081"
```
igalic: this is not the same as in the cache class

tests/conftest.py
gronke: Each test can decide whether to use the mocked Distribution module from the fixture or the imported one.

.cirrus.yml
```diff
@@ -0,0 +1,21 @@
...
+  freebsd_instance:
+    matrix:
+      #image: freebsd-11-2-release-amd64
+      image: freebsd-12-0-release-amd64
+  pip_cache:
+    folder: ~/.cache/pip
+    fingerprint_script: cat requirements*
+    populate_script: mkdir -p ~/.cache/pip
+  mirror_cache:
+    folder: .cache/libioc
```
igalic: Fingerprint is our version file.

tests/conftest.py
```diff
@@ -114,48 +97,64 @@ def logger() -> 'libioc.Logger.Logger':
...
+        logger=logger,
+        zfs=zfs
+    )
+    host = MockedHost(
+        datasets=datasets,
+        logger=logger,
+        zfs=zfs
     )
     return host
-    del host
```
igalic: very funny.

PR #584, opened 2018-12-23
MTU detection

gronke:
When no MTU was specified explicitly, the MTU of the host bridge (defined in `interfaces`) is detected.

python-iocage has implemented this feature in https://github.com/iocage/iocage/pull/722, but looks up the first bridge members MTU instead of the bridges value.

### DIFF
--- a/iocage/helpers_ioctl.py
+++ b/iocage/helpers_ioctl.py
@@ -1,0 +1,59 @@
+# Copyright (c) 2017-2018, Stefan Grönke
+# All rights reserved.
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted providing that the following conditions
+# are met:
+# 1. Redistributions of source code must retain the above copyright
+#    notice, this list of conditions and the following disclaimer.
+# 2. Redistributions in binary form must reproduce the above copyright
+#    notice, this list of conditions and the following disclaimer in the
+#    documentation and/or other materials provided with the distribution.
+#
+# THIS SOFTWARE IS PROVIDED BY THE AUTHOR ``AS IS'' AND ANY EXPRESS OR
+# IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
+# WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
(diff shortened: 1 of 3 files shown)


iocage/helpers_ioctl.py
```diff
@@ -0,0 +1,59 @@
...
+import ipaddress
+
+from enum import Enum
+
+
+class SOCKIO_IOCTLS(Enum):
+    """Hardcoded ioctl numbers."""
+
+    SIOCGIFADDR = -1071617759
+    SIOCGIFMTU = -1071617741
```
igalic: is there any reason you're not doing this thru `ctypes`, like this library does:

https://github.com/pydron/ifaddr/blob/master/ifaddr/_shared.py

?

gronke: Yep, tested in on all available FreeBSD and HardenedBSD variants and the values are unchanged and functional. If they break, then in new releases - but errors will be noticed quickly when starting the next jail with VNET.

gronke:
A quick test on performance:

### Popen ifconfig
```python3
import iocage.decorators

@iocage.decorators.timeit
def test_popen():
  import subprocess
  subprocess.Popen(
    ["ifconfig", "em0"],
    stdout=subprocess.PIPE,
    stderr=subprocess.PIPE
  ).communicate()

@iocage.decorators.timeit
def test_ioctl_helper():
  import iocage.helpers_ioctl
  iocage.helpers_ioctl.get_interface_ip4_address("em0")

def run(iterations):
  print("----- subprocess.Popen -----")
  [test_popen() for x in range(iterations)]
  print("--- iocage.helpers_ioctl ---")
  [test_ioctl_helper() for x in range(iterations)]

run(iterations=5)
```

```console
$ python3.6 speedtest.py
----- subprocess.Popen -----
function [test_popen] finished in 24.06597137451172 ms
function [test_popen] finished in 25.435209274291992 ms
function [test_popen] finished in 27.526378631591797 ms
function [test_popen] finished in 24.75738525390625 ms
function [test_popen] finished in 24.570941925048828 ms
--- iocage.helpers_ioctl ---
function [test_ioctl_helper] finished in 5.7430267333984375 ms
function [test_ioctl_helper] finished in 0.04482269287109375 ms
function [test_ioctl_helper] finished in 0.032901763916015625 ms
function [test_ioctl_helper] finished in 0.019550323486328125 ms
function [test_ioctl_helper] finished in 0.028848648071289062 ms
```

PR #586, opened 2018-12-28
Live-install of packages

gronke:
Packages can be installed and removed at runtime of a jail.

```sh
ioc create pkg-test
ioc start pkg-test
ioc pkg pkg-test vim-console
ioc exec pkg-test which vim
```

### DIFF
--- a/iocage/Pkg.py
+++ b/iocage/Pkg.py
@@ -34,6 +34,7 @@ import iocage.events
 import iocage.helpers
 import iocage.helpers_object
 import iocage.LaunchableResource
+import iocage.Config.Jail.File.Fstab
 
 _PkgConfDataType = typing.Union[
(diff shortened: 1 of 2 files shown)


iocage/Pkg.py
```diff
@@ -221,19 +224,25 @@ def install(
...
+                    command,
+                    passthru=False,
+                    event_scope=packageInstallEvent.scope
+                )
+                for event in jail_exec_events:
+                    if isinstance(event, iocage.events.JailLaunch) is True:
+                        if event.done is True:
+                            stdout = event.stdout.strip("\r\n")
+                    yield event
+            skipped = stdout.endswith("already installed")
```
igalic: i suppose this is quicker than checking

PR #593, opened 2018-12-29
Keep Pkg repository mounted

### LINKED ISSUE #592: Do not remove local package mirror from fstab
After installing packages the local package mirror mountpoint gets removed from the jails fstab file after finishing the installation. This is unfortunate because running `pkg update` within the jail fails when the iocage repository is enabled.

This regression was introduced with #586.

gronke:
fixes #592

- do not unmount iocage pkg repository after using Pkg in a jail

### DIFF
--- a/iocage/Pkg.py
+++ b/iocage/Pkg.py
@@ -229,7 +229,6 @@ class Pkg:
                 self.__mount_pkg_directory(jail)
                 stdout, stderr, code = jail.exec(["/bin/sh", "-c", command])
                 stdout = stdout.strip("\r\n")
-                self.__umount_pkg_directory(jail)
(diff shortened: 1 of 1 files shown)


iocage/Pkg.py
```diff
@@ -493,12 +492,3 @@ def __mount_pkg_directory(self, jail: iocage.Jail.JailGenerator) -> None:
             pass
         finally:
             jail.fstab.save()
-
-    def __umount_pkg_directory(self, jail: iocage.Jail.JailGenerator) -> None:
```
igalic: I'd keep the function around, might be useful to creating small surface containers

PR #594, opened 2018-12-30
Replace jail command fork with libjail-rs

gronke:
A work in progress branch of libjail-rs is providing Python bindings for its interfaces:

https://github.com/fubarnetes/libjail-rs/pull/14

With this bindings libiocage gains more control over jail interaction and also boosts performance.

### DIFF
--- a/iocage/Jail.py
+++ b/iocage/Jail.py
@@ -1617,85 +1617,77 @@ class JailGenerator(JailResource):
             return self.host.devfs[ruleset_line_position].number
 
     @property
-    def _launch_command(self) -> typing.List[str]:
+    def _launch_parameters(self) -> typing.Dict[str, typing.Union[str, int]]:
 
-        command = ["/usr/sbin/jail", "-c"]
+        import jail
+        stopped_jail = jail.StoppedJail(str(self.root_dataset.mountpoint))
+
+        p = dict()
 
         if self.config["vnet"]:
-            command.append("vnet")
+            p["vnet"] = 1
         else:
 
             if self.config["ip4_addr"] is not None:
                 ip4_addr = self.config["ip4_addr"]
-                command += [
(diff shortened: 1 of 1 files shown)


iocage/Jail.py
```diff
@@ -1683,34 +1678,30 @@ def _launch_parameters(self) -> typing.Dict[str, typing.Union[str, int]]:
...
             )
-            return stdout, stderr, returncode
+            raise e
+
+    def _launch_persistent_jail(
+        self,
+        passthru: bool
+    ) -> None:
+
+        running_jail = self._launch_jail()
```
fabianfreyer: It might make more sense to do this the other way around, and call `self._launch_persistent_jail()` from `_launch_jail()`, as `libjail-rs` jails are by default persistent, and you can drop persistence after spawning the command using `jail.RunningJail.defer_cleanup`.

iocage/Jail.py
```diff
@@ -1683,34 +1678,30 @@ def _launch_parameters(self) -> typing.Dict[str, typing.Union[str, int]]:
...
+        stopped_jail.parameters = self._launch_parameters
 
-        stdout, stderr, returncode = self._exec_host_command(
-            command=command,
-            passthru=passthru,
-            env=self.env
-        )
-        if returncode > 0:
+        try:
+            return jail.start()
```
fabianfreyer: this should be `stopped_jail.start()`

gronke:
Although it was tempting to use a well designed rust library, the idea was hold back in favor of [py-freebsd_sysctl](https://github.com/gronke/py-freebsd_sysctl) and [py-jail](https://github.com/gronke/py-jail) that are native Python implementations. The close relation to libjail-rs gives us the ability to share unit tests of those projects, so that either the Python and Rust bindings can benefit.

PR #608, opened 2019-01-10
Fix filtering templates

### LINKED ISSUE #607: template=no without effect
Release [v0.4.2](https://github.com/bsdci/libioc/releases/tag/0.4.2) has introduced a regression filtering templates from `ioc.Jails(["template=no"])`.

### Current Result
- templates appear in the list

### Expected Result
- the result does not iterate over templates

gronke:
fixes #607 

Parses `template` jail config property as bool, but stores as string `yes/no`.


gronke:
💜 I do, @igalic

PR #609, opened 2019-01-10
Sanitize internal JailConfig data types

gronke:
When setting an item on a JailConfig object and it is a plain property stored in JailConfig.data, the hardcoded defaults data type is looked up and used to sanitize the input.

- Prevents setting invalid values
- Improved error message when failing because of an invalid type
- Proper types for JailConfig properties

### DIFF
--- a/libioc/Config/Jail/BaseConfig.py
+++ b/libioc/Config/Jail/BaseConfig.py
@@ -27,6 +27,7 @@ import typing
 import re
 
 import libioc.Config.Data
+import libioc.Config.Jail.Defaults
 import libioc.Config.Jail.Properties
 import libioc.errors
 import libioc.helpers
@@ -617,7 +618,7 @@ class BaseConfig(dict):
                 setter_method(parsed_value)
                 return
 
-            self.data[key] = parsed_value
+            self.data[key] = self.__sanitize_value(key, parsed_value)
             error = None
         except ValueError as err:
             error = libioc.errors.InvalidJailConfigValue(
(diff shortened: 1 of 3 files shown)


ioc/Config/Jail/JailConfig.py
```diff
@@ -97,6 +98,45 @@ def __getitem__(self, key: str) -> typing.Any:
...
+    def __setitem__(  # noqa: T400
+        self,
+        key: str,
+        value: typing.Any,
+        skip_on_error: bool=False
+    ) -> None:
+        """
+        Normalize and set values to the configuration.
+
+        The type if the value defined in hardcoded defaults constraints the
```
igalic: the type *of* the

ioc/helpers.py
```diff
@@ -260,6 +260,37 @@ def parse_bool(data: typing.Optional[typing.Union[str, bool]]) -> bool:
...
+        TypeError: None is not a number
+        >>> parse_int("invalid")
+        TypeError: Value is not an integer: invalid
+        >>> parse_int(5.0)
+        5
+        >>> parse_int(5.1)
+        TypeError: Value is not an integer: 5.1
+    """
+    if data is None:
+        raise TypeError("None is not a number")
```
igalic: but `NaN` is, right? right??

gronke: NaN is not relevant for sanitization of Integer values, but it definitely is when sanitizing floats.

```python3
>>> float("nan").is_integer()
False
>>> import math
>>> math.isnan(float("nan"))
True
```

gronke: There are no float values possible in the configuration yet, which is why there is no parsing method for them.

PR #610, opened 2019-01-12
rename Python module to libioc

gronke:
Due to a name collision with another package called ioc, this module was renamed to `libioc`.

Please excuse the inconvenience caused by the recent fluctuation in the project and module names 😓 

This changes result in a bump of the minor version to `0.5.0`.


igalic:
can't believe we just lost these three additional char per line again 😾

PR #612, opened 2019-01-14
JailConfig parameter vnet must be boolean

gronke:
Before this change it was possible to store arbitrary strings as VNET value. Since this is a boolean configuration property, it must be boolean on set.


igalic:
do we not infer the type of a setting from its global default?

should this not have been caught after the merge of… oh, never mind, https://github.com/bsdci/libioc/pull/609 hasn't been merged yet 😅

PR #614, opened 2019-01-14
Accept non-absolute path as fstab source

### LINKED ISSUE #613: fstab lines without absolute path
When libioc parses an fstab line with a non-absolute source path, parsing the file fails.

```
$ ioc create gronke
$ echo "tmpfs /tmp tmpfs rw,mode=777 0 0" >> /iocage/jails/gronke/fstab
```

```python3
>>> libioc.Jail("gronke").fstab
  File "/usr/local/src/ioc/libioc/Jail.py", line 2206, in __getattribute__
    return object.__getattribute__(self, key)
  File "/usr/local/src/ioc/libioc/Jail.py", line 144, in fstab
    host=jail.host
  File "/usr/local/src/ioc/libioc/Config/Jail/File/Fstab.py", line 181, in __init__
    self.read_file()
  File "/usr/local/src/ioc/libioc/Config/Jail/File/Fstab.py", line 305, in read_file
    self._read_file_handle(f)
  File "/usr/local/src/ioc/libioc/Config/Jail/File/Fstab.py", line 319, in _read_file_handle
    self.parse_lines(f.read())
  File "/usr/local/src/ioc/libioc/Config/Jail/File/Fstab.py", line 290, in parse_lines
    "source": libioc.Types.AbsolutePath(source),
  File "/usr/local/src/ioc/libioc/Types.py", line 40, in __init__
    raise TypeError(f"Invalid value for AbsolutePath: {sequence}")
TypeError: Invalid value for AbsolutePath: tmpfs
```

### LINKED ISSUE #631: Obey rtsold_enable=NO sysrc
When starting a jail with `accept_rtadv` in any ip6_addr, ioc starts rtsold within the jail. It should be possible to opt out from that by explicitly setting `rtsold_enable="NO"` in the jails `/etc/rc.conf`.

Related to https://github.com/iocage/iocage/issues/711

gronke:
fixes #613

A Fstab source path does not necessarily require to be absolute (begin with a `/`).

https://github.com/iocage/iocage/issues/806 describes an fstab file that defines a tmpfs mountpoint within a jail. Even though this is a bad idea (because it bypasses jail resource limits and can lead to confusion of services within the jail), reading and creating fstab files should be possible using this library.

This changes enhance path verification and add the ability to use relative paths as fstab source path.

### DIFF
--- a/libioc/Config/Jail/File/Fstab.py
+++ b/libioc/Config/Jail/File/Fstab.py
@@ -68,8 +68,8 @@ class FstabLine(dict):
             self._escape("destination"),
             self.get("type", "nullfs"),
             self.get("options", "ro"),
-            self.get("dump", "0"),
-            self.get("passnum", "0")
+            str(self.get("freq", 0)),
+            str(self.get("passno", 0))
         ])
 
         if self["comment"] is not None:
@@ -92,7 +92,7 @@ class FstabLine(dict):
             dict.__setitem__(self, key, FstabFsSpec(value))
         elif key == "destination":
             dict.__setitem__(self, key, libioc.Types.AbsolutePath(value))
-        elif key in ["type", "options", "dump", "passnum", "comment"]:
+        elif key in ["type", "options", "freq", "passno", "comment"]:
             dict.__setitem__(self, key, value)
         else:
             raise KeyError(f"Invalid FstabLine key: {key}")
@@ -299,8 +299,8 @@ class Fstab(
                 "destination": libioc.Types.AbsolutePath(destination),
                 "type": fragments[2],
                 "options": fragments[3],
-                "dump": fragments[4],
-                "passnum": fragments[5],
+                "freq": int(fragments[4]),
+                "passno": int(fragments[5]),
                 "comment": comment
             })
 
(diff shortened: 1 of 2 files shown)


libioc/Config/Jail/File/Fstab.py
```diff
@@ -76,10 +76,16 @@ def __setitem__(
...
+            _type = libioc.Types.Path
+        elif key == "destination":
+            _type = libioc.Types.AbsolutePath
+
+        if _type is not None:  # source or destination
             if isinstance(value, str) is True:
-                absolute_path = libioc.Types.AbsolutePath(value)
-            elif isinstance(value, libioc.Types.AbsolutePath) is True:
+                absolute_path = _type(value)
+            elif isinstance(value, _type) is True:
```
igalic: what does this mean?

libioc/Types.py
```diff
@@ -27,20 +27,47 @@
 import re
 
 
-class AbsolutePath(str):
+class Path(str):
```
igalic: I'm not happy with this
i think we should be using https://docs.python.org/3/library/pathlib.html instead of string

libioc/Config/Jail/File/Fstab.py
gronke: This means that in case of `source`, either an instance of Path or AbsolutePath is accepted, in case of `destination`, only an AbsolutePath.

Anyhow, this logic can be moved to the Types.py, so that this code is less complex.

libioc/Types.py
gronke: We can address this suggestion in #617.

Issue #625, opened 2019-01-28
(help me) add provisioning method puppet


igalic:
Currently, i'm using `exec_start` to execute puppet on every startup, however, it seems it would be much easier if I could run it only once, at provisioning time.

I'd need some guidance to create such a provisioning plugin.

xaque208:
You can do something like `cat script.sh | iocage exec j sh` after the first start.  This is effectively what's happening in that jail module you worked on a while back.  Works pretty well in puppet, but the pattern could be applied outside well enough if something were orchestrating the jail build.

igalic:
it's turned out to be a little bit more elaborate… but for now we have #629

PR #629, opened 2019-02-03
Puppet (apply) provisioning plugin

igalic:
This patch introduces Puppet (apply) as provisioning method, addressing #625 

So far the design requires a (unique) name, a source (the control-repo) and an optional list of packages to be pre-installed.
By default that List of packages is `puppet6`, and if the source is a git repo, `rubygem-r10k`.

To begin the provisioning, we

- install puppet
- optionall install r10k
- (optionally) clone the control repo
- mount the control-repo
- (optionally, if the repo was cloned) run r10k
- run puppet

we could also consider to run puppet more than once, to guarantee idempotence.

### DIFF
--- a/libioc/Config/Jail/File/__init__.py
+++ b/libioc/Config/Jail/File/__init__.py
@@ -30,36 +30,6 @@ import libioc.helpers
 import libioc.helpers_object
 import libioc.LaunchableResource
 
-
-class ResourceConfig:
-    """Shared abstract code between various config files in a resource."""
-
-    def _require_path_relative_to_resource(
-        self,
-        filepath: str,
-        resource: 'libioc.LaunchableResource.LaunchableResource'
-    ) -> None:
-
-        if self._is_path_relative_to_resource(filepath, resource) is False:
-            raise libioc.errors.SecurityViolationConfigJailEscape(
-                file=filepath
-            )
-
-    def _is_path_relative_to_resource(
-        self,
-        filepath: str,
-        resource: 'libioc.LaunchableResource.LaunchableResource'
-    ) -> bool:
-
-        real_resource_path = self._resolve_path(resource.dataset.mountpoint)
-        real_file_path = self._resolve_path(filepath)
-
-        return real_file_path.startswith(real_resource_path)
-
-    def _resolve_path(self, filepath: str) -> str:
-        return os.path.realpath(os.path.abspath(filepath))
-
-
 class ConfigFile(dict):
     """Abstraction of UCL file based config files in Resources."""
 
@@ -221,7 +191,7 @@ class ConfigFile(dict):
         return None
 
 
-class ResourceConfigFile(ConfigFile, ResourceConfig):
+class ResourceConfigFile(ConfigFile):
     """Abstraction of UCL file based config files in Resources."""
 
     def __init__(
@@ -238,8 +208,5 @@ class ResourceConfigFile(ConfigFile, ResourceConfig):
     def path(self) -> str:
         """Absolute path to the file."""
         path = f"{self.resource.root_dataset.mountpoint}/{self.file}"
-        self._require_path_relative_to_resource(
-            filepath=path,
-            resource=self.resource
-        )
+        self.resource._require_relative_path(path)
         return os.path.abspath(path)
--- a/libioc/Provisioning/__init__.py
+++ b/libioc/Provisioning/__init__.py
@@ -28,6 +28,7 @@ import typing
 import libioc.errors
 import libioc.helpers
 import libioc.Provisioning.ix
+import libioc.Provisioning.puppet
 
 
 class Prototype:
--- a/libioc/Provisioning/__init__.py
+++ b/libioc/Provisioning/__init__.py
@@ -84,7 +84,8 @@ class Provisioner(Prototype):
 		self
 	) -> typing.Dict[str, Prototype]:
 		return dict(
-			ix=libioc.Provisioning.ix
+			ix=libioc.Provisioning.ix,
+			puppet=libioc.Provisioning.puppet
 		)
 
 	@property
--- a/libioc/Provisioning/__init__.py
+++ b/libioc/Provisioning/__init__.py
@@ -24,11 +24,90 @@
 # POSSIBILITY OF SUCH DAMAGE.
 """iocage provisioning prototype."""
 import typing
+import urllib.parse
 
 import libioc.errors
+import libioc.Types
 import libioc.helpers
 import libioc.Provisioning.ix
 import libioc.Provisioning.puppet
+
+_SourceType = typing.Union[
+    urllib.parse.ParseResult,
+    libioc.Types.AbsolutePath,
+]
+_SourceInputType = typing.Union[_SourceType, str]
+
+class Source(str):
+
+    _value: _SourceType
+
+    def __init__(
+        self,
+        value: _SourceInputType
+    ) -> None:
+       self.value = value
+
+    @property
+    def value(self) -> _SourceType:
+        return self._value
+
+    @value.setter
+    def value(self, value: _SourceInputType) -> None:
+
+        if isinstance(value, libioc.Types.AbsolutePath) is True:
+            self._value = typing.Cast(libioc.Types.AbsolutePath, value)
+            return
+        elif isinstance(value, urllib.parse.ParseResult) is True:
+            url = typing.Cast(urllib.parse.ParseResult, value)
+            self.__require_valid_url(url)
+            self._value = _value
+            return
+        elif isinstance(value, str) is False:
+            raise TypeError(
+                f"Input must be URL, AbsolutePath or str, but was {type(value)}"
+            )
+
+        try:
+            self._value = libioc.Types.AbsolutePath(value)
+            return
+        except ValueError as e:
+            pass
+
+        try:
+            url = urllib.parse.urlparse(value)
+            self.__require_valid_url(url)
+            self._value = url
+            return
+        except ValueError:
+            pass
+
+        raise ValueError("Provisioning Source must be AbsolutePath or URL")
+
+    @property
+    def local(self) -> bool:
+        """Return True when the source is local."""
+        return (isinstance(self.value, libioc.Types.AbsolutePath) is True)
+
+    @property
+    def remote(self) -> bool:
+        """Return True when the source is a remote URL."""
+        return (self.local is False)
+
+    def __require_valid_url(self, url: urllib.parse.ParseResult) -> None:
+        if url.scheme not in ("https", "http", "ssh", "git"):
+            raise ValueError(f"Invalid Source Scheme: {url.scheme}")
+
+    def __str__(self) -> str:
+        """Return the Provisioning Source as string."""
+        value = self.value
+        if isinstance(value, urllib.parse.ParseResult) is True:
+            return value.geturl()
+        else:
+            return str(value)
+
+    def __repr__(self) -> str:
+        return f"<Source '{self.__str__()}'>"
 
 
 class Prototype:
@@ -47,14 +126,14 @@ class Prototype:
 		self.__METHOD
 
 	@property
-	def source(self) -> typing.Optional[str]:
+	def source(self) -> typing.Optional[Source]:
 		config_value = self.jail.config["provisioning.source"]
-		return None if (config_value is None) else str(config_value)
+		return None if (config_value is None) else Source(config_value)
 
 	@property
 	def rev(self) -> typing.Optional[str]:
 		config_value = self.jail.config["provisioning.rev"]
-		return None if (config_value is None) else str(config_value)
+		return None if (config_value is None) else str(Source(config_value))
 
 	def check_requirements(self) -> None:
 		"""Check requirements before executing the provisioner."""
--- a/libioc/Provisioning/puppet.py
+++ b/libioc/Provisioning/puppet.py
@@ -114,7 +114,6 @@ class ControlRepoDefinition(dict):
                 "Source must be urllib.parse.ParseResult or absolute path"
             )
 
-
     @property
     def pkgs(self) -> typing.List[str]:
         """Return list of packages required for this Provisioning method."""
@@ -180,7 +179,7 @@ def provision(
     try:
         yield jailProvisioningAssetDownloadEvent.begin()
         pluginDefinition = ControlRepoDefinition(
-            url=urllib.parse.urlparse(self.source).geturl(),
+            source=urllib.parse.urlparse(self.source).geturl(),
             logger=self.jail.logger
         )
         yield jailProvisioningAssetDownloadEvent.end()
@@ -198,7 +197,7 @@ def provision(
         if not os.path.isdir(plugin_dataset_name):
             # only clone if it doesn't already exist
             git.Repo.clone_from(
-                pluginDefinition.url,
+                pluginDefinition.source,
                 plugin_dataset.mountpoint
             )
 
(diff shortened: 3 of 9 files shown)


libioc/Provisioning/puppet.py
```diff
@@ -0,0 +1,209 @@
+# Copyright (c) 2017-2019, Stefan Grönke, Igor Galić
+# Copyright (c) 2014-2018, iocage
```
gronke: This line can be removed.

libioc/Provisioning/__init__.py
```diff
@@ -84,7 +84,8 @@ def __available_provisioning_modules(
 		self
 	) -> typing.Dict[str, Prototype]:
 		return dict(
-			ix=libioc.Provisioning.ix
+			ix=libioc.Provisioning.ix,
+			puppet=libioc.Provisioning.puppet
```
igalic: we could split this to into `puppet-apply` and `puppet-server` (or `puppet-agent`?)

libioc/Provisioning/puppet.py
```diff
@@ -0,0 +1,195 @@
...
+        self,
+        name: str,
+        url: str,
+        logger: 'libioc.Logger.Logger'
+    ) -> None:
+        self.logger = logger
+        self.url = url
+        self.name = name
+
+        _pkgs = ['puppet6']  # make this a Global Varialbe
```
igalic: we can make the `6` a `provisioning.puppet.version=6` default.

libioc/Provisioning/puppet.py
```diff
@@ -0,0 +1,195 @@
...
+
+    try:
+        pkg = libioc.Pkg.Pkg(
+            logger=self.jail.logger,
+            zfs=self.jail.zfs,
+            host=self.jail.host
+        )
+
+        if os.path.isfile(f"{plugin_dataset.mountpoint}/post_install.sh"):
+            postinstall = ["/.puppet/post_install.sh"]
```
igalic: this doesn't exist in our puppet plugins…
we'll need to write our own.

libioc/ZFS.py
```diff
@@ -74,6 +74,27 @@ def get_or_create_dataset(
...
         return self.create_dataset(dataset_name)
 
+    def get_empty_dataset(
+        self,
+        dataset_name: str
+    ) -> libzfs.ZFSDataset:
+        """
+        Create and return an empty ZFS dataset.
+
+        Unlike get_or_create_dataset this method deletes existing datasets.
```
igalic: given how much effort this may be, and how we may want to reuse the same control repo, we might wanna actually use a persistent dataset here — ideas?

libioc/Provisioning/puppet.py
```diff
@@ -0,0 +1,195 @@
...
+        source=plugin_dataset.mountpoint,
+        destination="/.puppet",
+        options="ro",
+        auto_create_destination=True,
+        replace=True
+    )
+    self.jail.fstab.save()
+
+    if "pkgs" in controlRepoDefinition.keys():
+        pkg_packages = list(controlRepoDefinition["pkgs"])
```
gronke: pkgs is a property, not a a dict item. It needs to be accessed as `controlRepoDefinition.pkg`.

libioc/Provisioning/puppet.py
```diff
@@ -0,0 +1,195 @@
...
+    if not (self.source.startswith('file://') or self.source.startswith('/')):
+        # clone control repo
+        controlrepo_dataset_name = f"{self.jail.dataset.name}/puppet"
+        controlrepo_dataset = self.zfs.get_empty_dataset(
+            control_repo_dataset_name
+        )
+
+        git.Repo.clone_from(
+            ControlRepoDefinition["url"],
+            plugin_dataset.mountpoint
```
igalic: this variable does not exist, and i'll be honest

i'm having trouble choosing what to call this.

the confusion was between what call a "provisioning plugin", and an "iX plugin"

i really need some help with the our nomenclature here.

igalic: i think i've got it figured out…

libioc/ZFS.py
gronke: In this case just use `ZFS.get_or_create` instead and handle clone on init and pull in case it already exists.

igalic: pull would be handled by `r10k`.

gronke: In this case you can remove my commit moving the method to ZFS.py - it's really only used in the ix plugin.

igalic: done

libioc/Provisioning/puppet.py
```diff
@@ -0,0 +1,233 @@
...
+
+    try:
+        pkg = libioc.Pkg.Pkg(
+            logger=self.jail.logger,
+            zfs=self.jail.zfs,
+            host=self.jail.host
+        )
+
+        postinstall_script = pluginDefinition.generate_postinstall()
+        postinstall = "{self.jail.abspath}/launch-scripts/provision.sh"
```
igalic: i hope this is a sensible path

libioc/Provisioning/puppet.py
```diff
@@ -0,0 +1,233 @@
...
+                    basedir: /usr/local/etc/puppet/environments
+                    remote: {self.url}
+            >EOF
+
+            r10k deploy environment -p
+
+            """
+
+        postinstall += """
+        puppet apply /usr/local/etc/puppet/environments/manifests/site.pp
```
igalic: we could add `--debug` or similar if we knew the log level

libioc/Provisioning/puppet.py
```diff
@@ -0,0 +1,233 @@
...
+
+    @property
+    def pkgs(self) -> typing.List[str]:
+        """Return list of packages required for this Provisioning method."""
+        return self._pkgs
+
+    def generate_postinstall(self) -> str:
+        """Return list of strings representing our postinstall."""
+        postinstall = """#!/bin/sh
+        set -eu
```
igalic: we could also `set -x` here on appropriate log levels

libioc/Provisioning/puppet.py
```diff
@@ -148,27 +148,27 @@ def provision(
...
-        event_scope=_scope
+        scope=_scope
     )
 
     # download / mount provisioning assets
     try:
         yield jailProvisioningAssetDownloadEvent.begin()
         pluginDefinition = ControlRepoDefinition(
-            name=self.source.name,
+            name=self.name,
```
igalic: won't we have to add this to our DefaultConfig?

libioc/Provisioning/puppet.py
```diff
@@ -0,0 +1,233 @@
...
+            logger=self.jail.logger
+        )
+        yield jailProvisioningAssetDownloadEvent.end()
+    except Exception as e:
+        yield jailProvisioningAssetDownloadEvent.fail(e)
+        raise e
+
+    if pluginDefinition.remote:
+        mode = 'rw'  # we'll need to run r10k here..
+        plugin_dataset_name = f"{self.jail.dataset.name}/puppet"
```
igalic: given the effort we've put into allowing `provisioning.name`, it would be good to download plugins under their own dataset

libioc/Provisioning/puppet.py
```diff
@@ -54,27 +55,28 @@ class ControlRepoDefinition(dict):
...
-        name: str,
-        url: str,
+        url: urllib.parse.DefragResult,
         logger: 'libioc.Logger.Logger'
     ) -> None:
         self.logger = logger
+
+        if isinstance(url, urllib.parse.ParseResult) is False:
+            raise TypeError("Source must be a defragged URL")
         self.url = url
```
igalic: what is a `defragged` url?

libioc/Provisioning/puppet.py
igalic: query logger level here and set these

libioc/Provisioning/puppet.py
gronke: error message updated

libioc/Config/Jail/Defaults.py
```diff
@@ -99,7 +99,8 @@
     "provisioning": {
         "method": None,
         "source": None,
-        "rev": "master"
+        "rev": "master",
+        "name": None,
```
igalic: we should revert this again

libioc/Provisioning/__init__.py
```diff
@@ -55,6 +56,11 @@ def rev(self) -> typing.Optional[str]:
 		config_value = self.jail.config["provisioning.rev"]
 		return None if (config_value is None) else str(config_value)
 
+	@property
+	def name(self) -> typing.Optional[str]:
```
igalic: and this.

libioc/Config/Jail/Defaults.py
gronke: yeah, no name required

libioc/Provisioning/puppet.py
```diff
@@ -0,0 +1,237 @@
...
+    _pkgs: typing.List[str]
+
+    def __init__(
+        self,
+        url: urllib.parse.DefragResult,
+        logger: 'libioc.Logger.Logger'
+    ) -> None:
+        self.logger = logger
+
+        if isinstance(url, urllib.parse.ParseResult) is False:
```
igalic: so, this:
```sh-session
root@container-host1 /u/l/s/i/.libioc# ioc list -o name,provisioning.source webirc
+--------+-----------------------+
|  NAME  |  PROVISIONING.SOURCE  |
+========+=======================+
| webirc | /usr/local/etc/puppet |
+--------+-----------------------+
```

no longer works:

```python-traceback
root@container-host1 /u/l/s/i/.libioc# ioc -d spam provision webirc
/etc/rc.conf was read from /etc/rc.conf
Updated /etc/rc.conf data from /etc/rc.conf
Querying all running jails status
Executing: /usr/sbin/jls -v --libxo=json
    {"__version": "2", "jail-information": {"jail": [{"jid":15,"hostname":"webproxy","path":"/iocage/jails/webproxy/root","name":"default-webproxy","state":"ACTIVE","cpusetid":2, "ipv4_addrs": [], "ipv6_addrs": []}, {"jid":16,"hostname":"quassel","path":"/iocage/jails/quassel/root","name":"default-quassel","state":"ACTIVE","cpusetid":5, "ipv4_addrs": [], "ipv6_addrs": []}, {"jid":21,"hostname":"webirc","path":"/iocage/jails/webirc/root","name":"default-webirc","state":"ACTIVE","cpusetid":3, "ipv4_addrs": [], "ipv6_addrs": []}]}
    }
Injecting pre-loaded state to 'webirc'
[-] JailProvisioning@webirc: ...
  [-] JailProvisioningAssetDownload@webirc: FAILED [0.002s]
  File "/usr/local/bin/ioc", line 10, in <module>
    sys.dd:exit(cli())
  File "/usr/local/lib/python3.6/site-packages/click/core.py", line 722, in __call__
    return self.main(*args, **kwargs)
  File "/usr/local/lib/python3.6/site-packages/click/core.py", line 697, in main
    rv = self.invoke(ctx)
  File "/usr/local/lib/python3.6/site-packages/click/core.py", line 1066, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/usr/local/lib/python3.6/site-packages/click/core.py", line 895, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/usr/local/lib/python3.6/site-packages/click/core.py", line 535, in invoke
    return callback(*args, **kwargs)
  File "/usr/local/lib/python3.6/site-packages/click/decorators.py", line 17, in new_func                                                                                    
    return f(get_current_context(), *args, **kwargs)
  File "/usr/local/src/ioc/ioc_cli/provision.py", line 65, in cli
    **start_args
  File "/usr/local/src/ioc/ioc_cli/provision.py", line 102, in _provision
    print_function(_execute_provisioner(jail))
  File "/usr/local/src/ioc/ioc_cli/__init__.py", line 87, in print_events
    for event in generator:
  File "/usr/local/src/ioc/ioc_cli/provision.py", line 123, in _execute_provisioner
    for event in jail.provisioner.provision():
  File "/usr/local/lib/python3.6/site-packages/libioc/Provisioning/__init__.py", line 107, in provision                                                                      
    yield from self.__provisioning_module.provision(self)
  File "/usr/local/lib/python3.6/site-packages/libioc/Provisioning/puppet.py", line 182, in provision                                                                        
    raise e
  File "/usr/local/lib/python3.6/site-packages/libioc/Provisioning/puppet.py", line 177, in provision                                                                        
    logger=self.jail.logger
  File "/usr/local/lib/python3.6/site-packages/libioc/Provisioning/puppet.py", line 68, in __init__                                                                          
    raise TypeError("Source must be an URL")
TypeError: Source must be an URL
```
despite the code further down to support it.

libioc/Provisioning/puppet.py
gronke: Just be verbose here. The logger will only print the command output on verbose levels.

libioc/Provisioning/puppet.py
igalic: we should check for both AbsolutePath and URL

but we also have to catch whatever we throw here in the cli.

gronke: > but we also have to catch whatever we throw here in the cli.

Why does the CLI need to check that?

igalic: it needs to expect errors from the lib and print those, rather than a stack

gronke: Then there should be an `IllegalSourceError` that the library provides. Errors inheriting from `IocException` should cause the CLI to `exit(1)` rather than printing a stack trace.

igalic: this then should be the case for every error that happens at the edge of this library

gronke:
@igalic we still have to test remote repositories. Also it would be nice to provide authentication information for remote sources. Any ideas?

libioc/Config/Jail/File/__init__.py
```diff
@@ -30,36 +30,6 @@
 import libioc.helpers_object
 import libioc.LaunchableResource
 
-
-class ResourceConfig:
-    """Shared abstract code between various config files in a resource."""
```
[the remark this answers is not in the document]
gronke: The methods were moved to Resource itself. A resource should be able to check that a file is real relative to its storage location rather than File abstractions having this feature. You are right, the methods should be public.

libioc/Provisioning/puppet.py
```diff
@@ -0,0 +1,267 @@
...
+                logger=self.jail.logger
+            )
+        pluginDefinition = ControlRepoDefinition(
+            source=self.source,
+            logger=self.jail.logger
+        )
+        yield jailProvisioningAssetDownloadEvent.end()
+    except Exception as e:
+        yield jailProvisioningAssetDownloadEvent.fail(e)
+        raise e
```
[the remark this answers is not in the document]
gronke: Yes, they will bubble to the CLI. The CLI should not stack trace for exceptions inheriting from IocException, but here we forward any exception.

igalic:
@gronke wrote:
> @igalic we still have to test remote repositories. Also it would be nice to provide authentication information for remote sources. Any ideas?

so, right now, the repo i'm using is on gitlab and is only accessible with the correct SSH keys

however, it has *no* secretes and i could expose it publicly

as for authentication:

the easiest way would be to mount an (root's?) `~/.ssh/`?
It would also be nice if this could happen temporarily — i.e.: only during provisioning.

gronke:
> the easiest way would be to mount an (root's?) `~/.ssh/`?

Nein! We need proper key management to allow authenticated sources. How about an `.ssh` directory in a jails dataset (next to the config.json file) and the `provisioning.key=gronke` with `.ssh/gronke` and `.ssh/gronke.pub` as the key files.

PR #642, opened 2019-02-08
rename "provisioning.*" config namespace to "provision.*"

igalic:
this is also fits in well with the CLI command being `provision`!

### DIFF
--- a/libioc/Provisioning/__init__.py
+++ b/libioc/Provisioning/__init__.py
@@ -46,7 +46,7 @@ class Source(str):
         self,
         value: _SourceInputType
     ) -> None:
-       self.value = value
+        self._value = value
 
     @property
     def value(self) -> _SourceType:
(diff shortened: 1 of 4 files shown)


libioc/Provisioning/__init__.py
```diff
@@ -46,7 +46,7 @@ def __init__(
         self,
         value: _SourceInputType
     ) -> None:
-       self.value = value
+        self._value = value
```
gronke: `self.value` would process the input. The plain storage property should not be written here.

libioc/Provisioning/ix.py
```diff
@@ -22,7 +22,7 @@
 # STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING
 # IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
 # POSSIBILITY OF SUCH DAMAGE.
-"""iocage provisioner for ix-plugins."""
+"""ioc provisioner for ix-plugins."""
```
gronke: 👍

libioc/Provisioning/__init__.py
igalic: aye aye

PR #644, opened 2019-02-08
Unhide shared ZFS zvol devices

gronke:
When ZFS datasets are shared with a jail, their according zvol dev volumes should also appear in the jail.

For instance this is required when following the [article on VMs in jails](https://github.com/lattera/articles/blob/master/freebsd/2018-10-27_jailed_bhyve/article.md) by @lattera.

- Unhides the shared ZFS volumes and their child datasets
- Quotes dataset names in shell commands with shlex

### DIFF
--- a/libioc/Jail.py
+++ b/libioc/Jail.py
@@ -1609,6 +1609,25 @@ class JailGenerator(JailResource):
         if self._allow_mount_zfs == "1":
             devfs_ruleset.append("add path zfs unhide")
 
+        if self.config["jail_zfs"] is True:
+            unhidden_parents: typing.Set[str] = set()
+            shared_datasets = self._zfs_share_storage.get_zfs_datasets()
+            if len(shared_datasets) > 0:
+                devfs_ruleset.append("add path zvol unhide")
+                for shared_dataset in shared_datasets:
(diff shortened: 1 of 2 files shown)


libioc/ZFSShareStorage.py
```diff
@@ -58,23 +59,23 @@ def umount_zfs_shares(self) -> None:
                 "/sbin/zfs",
                 "set",
                 "jailed=off",
-                dataset.name
+                shlex.quote(dataset.name)
             ])
             self._exec([
                 "/sbin/zfs",
                 "umount",
```
igalic: i'm surprised we're not doing this with libzfs

libioc/Jail.py
```diff
@@ -1603,6 +1609,25 @@ def devfs_ruleset(self) -> libioc.DevfsRules.DevfsRuleset:
         if self._allow_mount_zfs == "1":
             devfs_ruleset.append("add path zfs unhide")
 
+        if self.config["jail_zfs"] is True:
```
igalic: do i understand the PR correctly that this can be both, a zfs dataset or a zvol?

PR #646, opened 2019-02-08
Obtain jail parameters from sysctl

gronke:
All jail parameters supported on a host can be obtained from sysctl. To do so this branch calls a native interface to get the param names, their type and maximum length.

~~The first draft obtains the data with py-sysctl, but a final version should use libc instead.~~

The prior py-sysctl Cython dependency has been replaced with a native Python 3 [freebsd_sysctl](https://github.com/gronke/py-freebsd_sysctl) module.

### DIFF
--- a/libioc/Config/Jail/Defaults.py
+++ b/libioc/Config/Jail/Defaults.py
@@ -94,7 +94,7 @@ DEFAULTS = libioc.Config.Data.Data({
     "mount_procfs": "0",
     "mount_devfs": "1",
     "mount_fdescfs": "0",
-    "securelevel": "2",
+    "securelevel": 2,
     "tags": [],
     "template": False,
     "jail_zfs": False,
--- a/libioc/Firewall.py
+++ b/libioc/Firewall.py
@@ -63,9 +63,10 @@ class Firewall:
             return
 
         try:
-            for key in self._required_sysctl_properties:
-                current = freebsd_sysctl.Sysctl(key).value
+            current = "not found"
+            for key in requirements:
                 expected = requirements[key]
+                current = freebsd_sysctl.Sysctl(key).value
                 if current == expected:
                     raise ValueError(
                         f"Invalid Sysctl {key}: "
--- a/libioc/Firewall.py
+++ b/libioc/Firewall.py
@@ -26,7 +26,7 @@
 import typing
 import shlex
 
-import sysctl
+import freebsd_sysctl
 
 import libioc.helpers
 import libioc.helpers_object
@@ -58,15 +58,28 @@ class Firewall:
     def ensure_firewall_enabled(self) -> None:
         """Raise an FirewallDisabled exception if the firewall is disabled."""
         requirements = self._required_sysctl_properties
-        requirement_keys = list(requirements.keys())
-        for item in sysctl.filter("net"):
-            if item.name in requirement_keys:
-                if item.value != requirements[item.name]:
-                    state = ("en" if (item.value == 0) else "dis") + "abled"
-                    raise libioc.errors.FirewallDisabled(
-                        hint=f"sysctl {item.name} is not {state}",
-                        logger=self.logger
+
+        if len(requirements) == 0:
+            return
+
+        try:
+            for key in self._required_sysctl_properties:
+                current = freebsd_sysctl.Sysctl(key).value
+                expected = requirements[key]
+                if current == expected:
+                    raise ValueError(
+                        f"Invalid Sysctl {key}: "
+                        f"{current} found, but expected: {expected}"
                     )
+            return
+        except Exception:
+            pass
+
+        hint = f"sysctl {key} is expected to be {expected}, but was {current}"
+        raise libioc.errors.FirewallDisabled(
+            hint=hint,
+            logger=self.logger
+        )
 
     def delete_rule(
         self,
--- a/libioc/Firewall.py
+++ b/libioc/Firewall.py
@@ -73,6 +73,7 @@ class Firewall:
                     )
             return
         except Exception:
+            # an IocageException is raised in the next step at the right level
             pass
 
         hint = f"sysctl {key} is expected to be {expected}, but was {current}"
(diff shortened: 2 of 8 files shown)


libioc/Host.py
```diff
@@ -187,8 +187,11 @@ def processor(self) -> str:
     @property
     def ipfw_enabled(self) -> bool:
         """Return True if ipfw is enabled on the host system."""
-        _sysctl = sysctl.filter("net.inet.ip.fw.enable")
-        return ((len(_sysctl) == 1) and (_sysctl[0].value == 1))
+        try:
+            firewall_enabled = freebsd_sysctl.Sysctl("net.inet.ip.fw.enable")
+            return (firewall_enabled.value == 1) is True
```
igalic: well, this…

libioc/Host.py
```diff
@@ -187,8 +187,11 @@ def processor(self) -> str:
     @property
     def ipfw_enabled(self) -> bool:
         """Return True if ipfw is enabled on the host system."""
-        _sysctl = sysctl.filter("net.inet.ip.fw.enable")
-        return ((len(_sysctl) == 1) and (_sysctl[0].value == 1))
```
igalic: …certainly reads much easier than this.

requirements.txt
```diff
@@ -1,2 +1,3 @@
 ucl==0.8.0
-gitpython
\ No newline at end of file
+gitpython
+freebsd_sysctl
```
igalic: yay!

gronke:
This affects the jail startup including single command execution. The jail parameters (and with that the usr.sbin/jail commands arguments) have been re-designed, so that we need to verify that all jail params are set as expected.

Also there is not a solution for swallowed jail parameters. When a host does not support a feature, the JailConfig is never asked for a value. This can happen because `security.jail.param` are looked up using [sysctl](https://github.com/gronke/py-freebsd_sysctl), where unsupported parameters do not appear. Certain ioc features require sysctl status checks that throw an exception in case of missing host features.

libioc/JailParams.py
```diff
@@ -0,0 +1,180 @@
...
+
+    __base_class = freebsd_sysctl.Sysctl
+
+    def __setitem__(self, key: str, value: typing.Any) -> None:
+        """Set of jail params sysrc is not supportes."""
+        raise NotImplementedError("jail param sysctl cannot be modified")
+
+    def __delitem__(self, key: str) -> None:
+        """Delete of jail param sysrc not supported."""
+        raise NotImplementedError("jail param sysctl cannot be deleted")
```
igalic: what's up with "sysrc" in the comments vs "sysctl" in the exceptions here?

libioc/Jail.py
```diff
@@ -1647,93 +1648,66 @@ def devfs_ruleset(self) -> libioc.DevfsRules.DevfsRuleset:
...
+                    # vnet is only used when explicitly enabled
+                    # (friendly to Kernels without VIMAGE support)
+                    continue
+                value = "vnet"
+            else:
+                config_property_name = sysctl.iocage_name
+                if self.config._is_known_property(config_property_name):
+                    value = config[config_property_name]
+                else:
+                    continue
```
igalic: continue?
isn't this some error case?

libioc/Firewall.py
```diff
@@ -58,15 +58,28 @@ def _required_sysctl_properties(self) -> typing.Dict[str, int]:
...
+                expected = requirements[key]
+                if current == expected:
+                    raise ValueError(
+                        f"Invalid Sysctl {key}: "
+                        f"{current} found, but expected: {expected}"
                     )
+            return
+        except Exception:
+            pass
+
```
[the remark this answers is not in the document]
gronke: The Exception is raised in the lines below. By leaving the above exception, the upcoming Exception stack is not unnecessarily deep. The try/except is used to check for the existence of the given Sysctl.

libioc/Jail.py
gronke: Yes, the loop iteration ends here.

libioc/JailParams.py
```diff
@@ -0,0 +1,180 @@
...
+        """Return the name of the param formatted for the jail command."""
+        name = str(self.name)
+        prefix = "security.jail.param."
+        if name.startswith(prefix) is True:
+            return name[len(prefix):]
+        return name
+
+    @property
+    def iocage_name(self) -> str:
+        """Return the name of the param formatted for iocage config."""
```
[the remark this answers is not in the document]
gronke: Not yet, but it will be accessed through this property. (Future changes become easier)

libioc/Firewall.py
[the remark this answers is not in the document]
igalic: i suggest adding this as comment at the beginning of the try

libioc/Config/Jail/Defaults.py
```diff
@@ -94,7 +94,7 @@
     "mount_procfs": "0",
     "mount_devfs": "1",
     "mount_fdescfs": "0",
-    "securelevel": "2",
+    "securelevel": 2,
```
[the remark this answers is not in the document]
gronke: The `mount_*` BaseConfig items relate to usr.sbin/jail arguments, not to jail params. Handling of them remains unchanged, so that a change is not required in terms of introducing py-freebsd_sysctl

libioc/Jail.py
```diff
@@ -1653,93 +1654,70 @@ def devfs_ruleset(self) -> libioc.DevfsRules.DevfsRuleset:
...
+
+            if sysctl_name == "security.jail.param.devfs_ruleset":
+                value = str(self.devfs_ruleset)
+            elif sysctl_name == "security.jail.param.path":
+                value = self.root_dataset.mountpoint
+            elif sysctl_name == "security.jail.param.name":
+                value = self.identifier
+            elif sysctl_name == "security.jail.param.allow.mount.zfs":
+                value = str(self._allow_mount_zfs)
+            elif sysctl_name == "security.jail.vnet":
```
igalic: what's the difference between `security.jail.param.*` and `security.jail.*`?

gronke: Some keys are duplicated in `security.jail.*` and `security.jail.param.*`, but we only handle the latter. Fixed by rebasing.

PR #647, opened 2019-02-08
Release 0.6.0

gronke:
## Breaking
- The Jail Config namespace `provisioning` was renamed to `provision`

## Fixes
- FreeBSD 12 release updates by using a virtual basedir for freebsd-update

## Enhancements
- Validate BaseConfig  when cloning
- Pass through ZFS /dev/zvol devices when sharing ZFS datasets with a jail

## Features
- Bhyve VMs in Jails (`allow.vmm`)
- Puppet provisioning of jails from local and remote control repos


igalic:
breaking changes are missing

gronke:
@igalic good catch. The body is updated!

PR #653, opened 2019-02-11
check hostid on Jail.start when hostid_strict_check is enabled

### LINKED ISSUE #634: Host ID
At the current time there is no check implemented that validates a jail's hostid. Please leave a comment if you find a use-case for that, so that the feature can be added to ioc.

gronke:
closes #634 

When the JailConfig property `hostid_strict_check` is enabled, the jails `hostid` is matched against `/etc/hostid` on the host. libioc refuses to start Jails with mismatch in the values with a `JailHostIdMismatch` exception.

### DIFF
--- a/libioc/Host.py
+++ b/libioc/Host.py
@@ -55,6 +55,7 @@ class HostGenerator:
     _devfs: libioc.DevfsRules.DevfsRules
     _defaults: libioc.Resource.DefaultResource
     _defaults_initialized = False
+    __hostid: str
     releases_dataset: libzfs.ZFSDataset
     datasets: libioc.Datasets.Datasets
     distribution: _distribution_types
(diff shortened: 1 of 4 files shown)


libioc/Host.py
```diff
@@ -114,6 +115,19 @@ def _init_defaults(
                 zfs=self.zfs
             )
 
+    @property
+    def id(self) -> str:
```
gronke: This should be `uuid.UUID`, but the uuid module in Python 3.6 is loading terribly slow.

igalic: I thought we fixed that ourselves somewhere?

gronke: Yes, but it landed in Python 3.7, which is currently unsupported due to py-libzfs (see the ports Makefile).

igalic: we could add the shim back in, but only load it on < 3.7

PR #654, opened 2019-02-11
existing data keys do not cause UnknownJailProperty when read

gronke:
Fixes an issue interfacing with legacy jails that have unsupported jail properties. When reading such values from jails that have loaded them from an existing configuration storage (json, zfs, ucl), no error is raised. This allows to unset such values and starting jails with existing, but unsupported JailConfig parameters.

### DIFF
--- a/libioc/Jail.py
+++ b/libioc/Jail.py
@@ -589,6 +589,14 @@ class JailGenerator(JailResource):
             self.storage_backend.apply(self.storage, self.release)
 
         if quick is False:
+            unknown_config_parameters = list(
+                self.config.unknown_config_parameters
+            )
+            if len(unknown_config_parameters) > 0:
+                _unused_parameters = str(", ".join(unknown_config_parameters))
+                self.logger.warn(
+                    f"Unused JailConfig parameters: {_unused_parameters}"
+                )
             self._save_autoconfig()
 
         try:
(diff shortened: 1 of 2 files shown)


gronke:
This draft comes with the disadvantage of no warnings being printed when starting a jail with unknown properties. Either we have to double-check whether a unused property was found or the idea to assume existing keys of the JailConfig.data property are known was thought too simple.

gronke:
Added a detection step for unused jail parameters in `JailGenerator.start()` that prints one warning for all unused JailConfig parameters found in an existing Jail on start.

igalic:
why is travis so unhappy with this?

libioc/Jail.py
```diff
@@ -589,6 +589,14 @@ def _stop_failed_jail(
             self.storage_backend.apply(self.storage, self.release)
 
         if quick is False:
+            unknown_config_parameters = list(
+                self.config.unknown_config_parameters
+            )
+            if len(unknown_config_parameters) > 0:
+            _unused_parameters = str(", ".join(unknown_config_parameters))
```
igalic:
```
./libioc/Jail.py:596:13: E112 expected an indented block
./libioc/Jail.py:596:30: E999 IndentationError: expected an indented block
```

libioc/Jail.py
```diff
@@ -589,6 +589,14 @@ def _stop_failed_jail(
...
+                self.logger.warn(
```
igalic:
```
./libioc/Jail.py:597:17: E113 unexpected indentation
```

PR #661, opened 2019-02-17
remove ip addresses after stopping a jail

### LINKED ISSUE #659: IP is not removed on destroy
If you issue `ioc destroy myjail` IP alias is left as it is.

```
ioc create myjail ip4_addr="lo1|172.16.0.10"
ioc start myjail
ifconfig lo1 | grep 172.16.1.10
ioc stop myjail
ioc destroy myjail
ifconfig lo1 | grep 172.16.1.10
```
After that IP still attached to the interface. Using version **ioc, version 0.6.0 2019/02/08**

gronke:
closes #659 

- remove any configured IPv4 or IPv6 address from non-vnet jails in exec_poststop hook

### DIFF
--- a/libioc/Jail.py
+++ b/libioc/Jail.py
@@ -890,11 +890,9 @@ class JailGenerator(JailResource):
             exec_prestop.append(self.config["exec_prestop"])
         if self.config["exec_stop"] is not None:
             exec_stop.append(self.config["exec_stop"])
+        exec_poststop = self._stop_network() + exec_poststop
         if self.config["exec_poststop"] is not None:
             exec_poststop.append(self.config["exec_poststop"])
-
-        if self.config["vnet"]:
-            exec_poststop = self._stop_vimage_network() + exec_poststop
 
         if self.config["jail_zfs"] is True:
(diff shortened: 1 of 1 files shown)


gronke:
@urosgruber with this changes your IP addresses should be removed from the configured interfaces as expected. Review welcome ☺️

libioc/Jail.py
```diff
@@ -1940,6 +1938,31 @@ def _start_vimage_network(self) -> typing.Tuple[
...
+            config_value = self.config[f"ip{protocol}_addr"]
+            if config_value is None:
+                return
+            for nic, addresses in config_value.items():
+                if addresses is None:
+                    continue
+                for address in addresses:
+                    if isinstance(address, str):
+                        # skip DHCP and ACCEPT_RTADV
+                        continue
```
igalic: why?

urosgruber:
Just tested and looks like it's working fine.

libioc/Jail.py
gronke: They are not accepted by jail during start. So it must be a mistake that should not prevent the teardown.

PR #663, opened 2019-02-19
utilise py-jail instead of a cached jail state

gronke:
- replaces fork of `/usr/sbin/jls` to query a jails JID
- [py-jail](https://github.com/gronke/py-jail) is a libc binding for jail interaction using native Python3
- Always queries the jail live with libc.jail_get - no more need for manual state querying

- breaks ioc CLI commans that expect the jail.state object


gronke:
@himrock922 with this you no longer have to call `jail.state.query()` manually!

Again, thanks to @fabianfreyer, who brought the inspiration this step with [libjail-rs](https://github.com/fubarnetes/libjail-rs/).

libioc/Jail.py
```diff
@@ -1640,7 +1640,7 @@ def _launch_args(self) -> typing.List[str]:
                     # vnet is only used when explicitly enabled
                     # (friendly to Kernels without VIMAGE support)
                     continue
-                value = "vnet"
+                value = 1
```
igalic: oops

gronke:
The features used by libioc are now [unit-tested in py-jail](https://github.com/gronke/py-jail/commit/251a3fc2e2421189339e1ce93dffc0581b7df3c1).

tests/conftest.py
```diff
@@ -175,6 +178,49 @@ def release(
...
+    del release
+
+
+@pytest.fixture(scope="function")
+def new_jail(
+    host: 'libioc.Host.Host',
+    logger: 'libioc.Logger.Logger',
+    zfs: libzfs.ZFS,
+) -> 'libioc.Jail.Jail':
+    jail_name = "new-jail-" + str(random.randint(1, 32768))
```
igalic: this might be better done with Hypothesis

tests/conftest.py
```diff
@@ -175,6 +178,49 @@ def release(
...
+        logger=logger,
+        zfs=zfs
+    )
+    yield new_jail
+    new_jail.destroy(force=True)
+
+
+@pytest.fixture(scope="function")
+def bridge_interface() -> str:
+    bridge_name = "bridgeTest" + str(random.randint(1024, 4096))
```
igalic: again, this might be better done with Hypothesis: https://hypothesis.readthedocs.io/en/latest/data.html

PR #668, opened 2019-02-21
Fix docs path and build

gronke:
- The paths configured in Sphinx were updated form `ioc` to `libioc`
- Circular dependency between BaseConfig / Defaults resolved


gronke:
https://bsdci.github.io/libioc/index.html updated

PR #630, opened 2019-02-03
add config with no exists config

himrock922:
I commit of parameter that was said "unknown" with config.
For example, `The config property 'vnet_default_interface' is unknown`
Because I don't know very well about config detail, I need refactoring it.

This PR addresses #626

Issue #636, opened 2019-02-03
cpuset


gronke:
Assigning a jail to a specific CPU is related to Resource Limits.

Please see the [cpuset man page](https://www.freebsd.org/cgi/man.cgi?query=cpuset&sektion=1&apropos=0&manpath=FreeBSD+12.0-RELEASE+and+Ports)

It is important to solve #413 before implementing this feature to come up with a consistent interface. Moving all Resource Limits to their own namespace will make clear what those parameters (and cpuset) do.

Issue #632, opened 2019-02-03
Track last_started date of jails


gronke:
It is reasonable to track when a jail was last started successfully. Unlike iocage this should not be stored in a Jail configuration.

igalic:
this is something an external orchestration tool would have a much easier time tracking. and a much more understandable need for doing it to begin with

himrock922:
If the orchestration tool makes track, Should libioc returns only response with last_started?

https://github.com/bsdci/libioc/blob/master/libioc/Jail.py#L2289

If above line changes that like return response datetime for last_started, jail config  not store and return last_started...?

gronke:
We currently fork a process to query jail state. In Python this is an expensive operation, so that we first need to create interfaces for libc jail_get and jail_set. 

@fabianfreyer has pioneered on a Rust library with Python interface that is about to be utilized in #594. I'm not so much in favor to depend on Rust, but after a review, it is a well designed library.

With this we will be able to always query jail state on the fly at low cost, so that changed state can easily be signaled from changing operations (I'm thinking about the JailEvent informing about state change).

PR #671, opened 2019-02-23
add config  with no exists config

himrock922:
ref #630
```
I commit of parameter that was said "unknown" with config.
For example, `The config property 'vnet_default_interface' is unknown`
Because I don't know very well about config detail, I need refactoring it.

This PR related to issue #626
```

Hi All,
I'm sorry, after rebase execute, confflict code(like as <<<HEAD, ======) mixed PR(Also, reflog failed). So, recreate branch...

Make sure to follow and check these boxes before submitting a PR! Thank you.

- [x] Explain the feature
- [x] Read [CONTRIBUTING.md](https://github.com/bsdci/libioc/blob/master/CONTRIBUTING.md)


gronke:
> Hi All,
I'm sorry, after rebase execute, confflict code(like as <<<HEAD, ======) mixed PR(Also, reflog failed). So, recreate branch...

No worries, your set of changes is great to spot the differences to other variants.

Adding the values to the defaults has the effect that ioc does no longer complain about unknown JailConfig properties, but they will just be ignored when processing a jail. There are the cases:

1. A feature is missing in libioc
    1. feature needs to be implemented
    2. config property can be added to Globals
2. An iocage feature is unfortunately named
    1. the value can be mapped to a more reasonable internal name
    2. mapped name is added to the Globals
3. iocage duplicates a feature with other name
    1. ignore the value on read (maybe verify with the other settings, but still ignore it)
    2. calculate the value on configuration write, so that the result remains compatible
3. The iocage feature has design flaws
    1. will not be implemented
    2. warn users with error
    3. a `--ignore-compat` flag (or similar) can turn the error (2) into a warning and continue

As one with the ideas from your other branch, @himrock922, I'd like to derive more Issues from each individual change you suggest to the Globals. Doing so, while following the workflow described above, will lead to a user friendly solution with minimal breakage for iocage users. (iocage_legacy should not be affected).

libioc/Config/Jail/Globals.py
```diff
@@ -67,7 +67,7 @@
     "allow_mlock": None,
     "allow_mount": 0,
     "allow_mount_devfs": 0,
-    "allow_mount_fusefs": 0,
+    "allow_mount_fusefs": False,
```
gronke: Interesting idea to actually turn the `allow_mount_*` properties into booleans. The sysctl value is an integer type, but there is a translation of boolean values happening. We should just be consistent and change similar properties among this one.

libioc/Config/Jail/Globals.py
```diff
@@ -26,6 +26,7 @@
 
 
 DEFAULTS = libioc.Config.Data.Data({
+    "CONFIG_VERSION": 17,
```
gronke: libioc does not read the `CONFIG_VERSION` at all, so should we not put it in the defaults, but allow this specific property to be ignored when found to be existing? Setting or changing it should not be possible and fail with the usual UnknownConfigProperty exception.

libioc/Config/Jail/Globals.py
```diff
@@ -61,16 +64,28 @@
...
     "allow_mount_tmpfs": 0,
     "allow_quotas": 0,
     "allow_socket_af": 0,
+    "allow_tun": 0,
     "allow_vmm": False,
+    "available": 0,
+    "bpf": None,
+    "comment": None,
+    "compression": None,
+    "compressratio": None,
```
gronke: This is not a property that can be set. It should not be in the jail configuration at all. Is that the case in python-iocage?

libioc/Config/Jail/Globals.py
```diff
@@ -61,16 +64,28 @@
...
+    "count": None,
+    "cpuset": False,
```
gronke: There is a new feature suggested in #636 - this change should be handled in a separate PR that addresses the GitHub Issue.

libioc/Config/Jail/Globals.py
```diff
@@ -84,20 +99,36 @@
...
     "jail_zfs": False,
     "jail_zfs_dataset": None,
+    "jail_zfs_mountpoint": None,
     "provision": {
         "method": None,
         "source": None,
         "rev": "master"
-    }
+    },
+    "last_started": None
```
gronke: Missing feature tracked in #632 - this change belongs in a separate PR.

gronke (changes requested):
Hi @himrock922, the comments I left are a start to implement your suggested changes. Some changes require new features to be implemented, which you can find linked in my comments. If you would create one PR per feature, we can individually discuss and test the implications for libioc.

libioc/Config/Jail/Globals.py
himrock922: `but allow this specific property to be ignored when found to be existing?`
I think unnecessary it.
Maybe, If it did implement, I think that code difficult to read.(If not used, that should be remove.)
Next commit, I will delete `CONFIG_VERSION`.
Regards.

libioc/Config/Jail/Globals.py
himrock922: OK. I think should keep a consistency for code(For example, 0 or 1 integer only used, or boolean type only used).
If possible, To refactoring for future, I think should only use boolean type.

himrock922: Anyway, I do fix such `allow_mount.*` parameters from 0,1 to True, False.

libioc/Config/Jail/Globals.py
himrock922: Yes,  such parameters using with python-iocage.
https://www.freebsd.org/cgi/man.cgi?query=iocage&sektion=8

* available
https://www.freebsd.org/cgi/man.cgi?query=iocage&sektion=8

```
     available	   Available space in the jail's dataset.  The amount of space
		   available to	the dataset and	all its	children, assuming
		   that	there is no other activity in the pool.	 Because space
		   is shared within a pool, availability can be	limited	by any
		   number of factors, including	physical pool size, quotas,
		   reservations, or other datasets within the pool.

		   Source: zfs(8)
```

* bpf

https://github.com/iocage/iocage/blob/528f007e92f6e081614d8a09186c4831fc9f3e32/iocage.8#L1019

* compression

https://github.com/iocage/iocage/blob/528f007e92f6e081614d8a09186c4831fc9f3e32/iocage.8#L1724

```
Controls the compression algorithm used for this dataset.
The lzjb compression algorithm is optimized for performance while
providing decent data compression.
Setting compression to on uses the lzjb compression algorithm.
The gzip algorithm uses the same compression as the
```

* compressratio

https://github.com/iocage/iocage/blob/528f007e92f6e081614d8a09186c4831fc9f3e32/iocage.8#L1789

```
Compression ratio.
Read-only.
For non-snapshots, the compression ratio achieved for the used space
of this dataset, expressed as a multiplier.
The used property includes descendant datasets, and, for clones, does
not include the space shared with the origin snapshot.
```

Also, Also, `comment` property setting a lower layer for `user` property.

ref https://github.com/bsdci/libioc/pull/630#discussion_r253304269

For example
```
user {
  comment: None,
  notes: None
}

PR #674, opened 2019-02-23
Release 0.6.1

gronke:
## Features
- Check `hostid` when `hostid_strict_check` is enabled

## Enhancements
- Use native Python [py-freebsd_sysctl](https://github.com/gronke/py-freebsd_sysctl) libc bindings to read supported `security.jail.param` sysctls
- Lookup jail state (jid) with [py-jail](https://github.com/gronke/py-jail)
- Always query jail state when accessing the JailGenerator.running property, but de-bounce JailGenerator.jid
- remove IP addresses from non-VNET interfaces after stopping a jail

## Fixes
- Add missing ix-plugins provisioner errors
- Fix fetching of packets for releases not matching the host release (via ABI= env variable)
- Allow backup importer to read nested JSON config

## Chores
- Test automation with Cirrus Ci (pass is mandatory merge criteria)
- Release and Update mirror cache for testing with asset cache
- Refactor jail parameters into JailParams class


igalic:
> - Use native Python py-freebsd_sysctl libc bindings to read supported `security.jail.param` sysctls
> - Lookup jail state (jid) with py-jail

these aren't just enhancements, they are also breaking changes / new dependencies.

this is should be 0.7.0

gronke:
Replaced by #675 after [noticing](https://github.com/bsdci/libioc/pull/674#issuecomment-466690112) the breaking changes.

PR #675, opened 2019-02-23
Release 0.7.0

gronke:
## Breaking
- The JailState is no longer available as JailGenerator.state
  - `Jail.state["jid"]` was the only item used, which is now queried with [py-jail](https://github.com/gronke/py-jail) transparently
- ZFS basejail datasets are no longer created by default
  - Setting `ioc_legacy_support="YES"` in /etc/rc.conf re-enables the compatibility feature

## Features
- Check `hostid` when `hostid_strict_check` is enabled
- Support for `$http_proxy` env variable for fetching releases and updates

## Enhancements
- Use native Python [py-freebsd_sysctl](https://github.com/gronke/py-freebsd_sysctl) libc bindings to read supported `security.jail.param` sysctls
- Lookup jail state (jid) with [py-jail](https://github.com/gronke/py-jail)
- Always query jail state when accessing the JailGenerator.running property, but de-bounce JailGenerator.jid
- remove IP addresses from non-VNET interfaces after stopping a jail

## Fixes
- Add missing ix-plugins provisioner errors
- Fix fetching of packets for releases not matching the host release (via ABI= env variable)
- Allow backup importer to read nested JSON config
- Handling of boolean value jail_zfs is fixed and consistent

## Chores
- Test automation with Cirrus Ci (pass is mandatory merge criteria)
- Release and Update mirror cache for testing with asset cache
- Refactor jail parameters into JailParams class


igalic:
I'm confused. how come tests on #672 were failing, but they're okay here, with no changes in between

gronke:
> I'm confused. how come tests on #672 were failing, but they're okay here, with no changes in between

For some reason Cirrus Ci triggers another build when a branch is merged. Let's observe it, but it's important in first place (an mandatory merge criteria for the master branch) to pass the tests in PRs.

PR #686, opened 2019-03-20
raise on access of an unknown default jail config property

gronke:
Helps resolving an issue in the ioc CLI tool reporting access to unknown jail config properties.


igalic:
maybe a test-case for this could be useful?

gronke:
An upcoming PR will cover this in findings, together with more extensive tests of JailConfig, DefaultConfig and global (hardcoded) defaults.

PR #689, opened 2019-03-21
Fix VNET MAC address setting

### LINKED ISSUE #688: Setting custom MAC address fails on the first time
The first time a dynamic <NIC>_mac property is set, an UnknownConfigProperty exception is raised. When the value already exists, the MAC address can be set without this message.
 
```console
$ ioc set vnet0_mac=63ECAC12D0F1 testmac         
The config property 'vnet0_mac' is unknown
Jail 'foobar2' updated: vnet0_mac
$ ioc set vnet0_mac=63ECAC12D0F2 testmac
Jail 'foobar2' updated: vnet0_mac
```

One problem is that an error should interrupt the execution (so the value should *not* be updated after the message), another is that on set any <NIC>_mac config property is valid. It would be possible to enforce the according interface to exist, but this would impact the order of setting them:

```sh
ioc create testmac1 vnet=on interfaces="vnet0:bridge0" vnet0_mac=63ECAC12D0F1
ioc create testmac1 vnet=on vnet0_mac=63ECAC12D0F1 interfaces="vnet0:bridge0"
```

A solution to work around this issue is to apply config properties in order of priority.

Thanks to @leonbayer for reporting this issue.

gronke:
fixes #688 

- do not print error message when setting `<NIC>_mac` for the first time
- be strict when setting VNET interface mac addresses, but permissive when reading config files

### DIFF
--- a/libioc/Config/Jail/BaseConfig.py
+++ b/libioc/Config/Jail/BaseConfig.py
@@ -823,7 +823,7 @@ class BaseConfig(dict):
         special_properties = libioc.Config.Jail.Properties.properties
         return list(properties.union(special_properties))
 
-    def _key_is_mac_config(self, key: str, explicit: bool=False) -> bool:
+    def _key_is_mac_config(self, key: str, explicit: bool=True) -> bool:
         fragments = key.rsplit("_", maxsplit=1)
         if len(fragments) < 2:
             return False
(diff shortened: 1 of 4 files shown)


igalic:
do we have other such *dynamically named properties*, or are macs the only one?

gronke:
> do we have other such *dynamically named properties*, or are macs the only one?

Only `*_mac` and `user.*` are dynamically named config properties.

PR #690, opened 2019-03-24
Start jails with libc jail_set

gronke:
- Refactoring of the entire start process
  - Individual events for actions and hooks
- Use py-jail to start jails
- (Un)mount NullFS and DevFS with libc
- Deprecate launch helper scripts (`/.iocage` in a jail)
- Do not store basejail directories in a jails fstab file

### DIFF
--- a/libioc/Config/Jail/File/Fstab.py
+++ b/libioc/Config/Jail/File/Fstab.py
@@ -34,6 +34,7 @@ import libioc.helpers_object
 import libioc.Types
 import libioc.Config.Jail
 import libioc.Config.Jail.File
+import libioc.Storage.Basejail
 
 
 class FstabFsSpec(libioc.Types.AbsolutePath):
@@ -558,24 +559,6 @@ class JailFstab(Fstab):
             file=file
         )
 
-    @property
-    def maintenance_lines(self) -> typing.List[FstabMaintenanceLine]:
-        """Auto-generate lines that are required for jail start and stop."""
-        if self.jail is None:
-            return []
-
-        return [FstabMaintenanceLine(dict(
-            source=libioc.Types.AbsolutePath(self.jail.launch_script_dir),
-            destination=libioc.Types.AbsolutePath(
-                f"{self.jail.root_dataset.mountpoint}/.iocage"
-            ),
-            options="ro",
-            type="nullfs",
-            freq=0,
-            passno=0,
-            comment=self.AUTO_COMMENT_IDENTIFIER
-        ))]
-
     def parse_lines(
         self,
         input_text: str,
@@ -596,58 +579,18 @@ class JailFstab(Fstab):
             exclude_destinations:
                 List of destination strings that is skipped
         """
-        # skip lines with destinations overlapping self.basejail_lines
-        skip_destinations += list(map(
-            lambda x: str(x['destination']),
-            self.basejail_lines
-        ))
+        BasejailStorage = libioc.Storage.Basejail.BasejailStorage
+        if isinstance(self.jail.storage_backend, BasejailStorage) is True:
+            skip_destinations += list(map(
+                lambda x: str(x[1]),
+                self.jail.storage_backend.basejail_mounts
+            ))
         Fstab.parse_lines(
             self,
             input_text=input_text,
             ignore_auto_created=ignore_auto_created,
             skip_destinations=skip_destinations
         )
-
-    @property
-    def basejail_lines(self) -> typing.List[FstabBasejailLine]:
-        """
-        Auto-generate lines of NullFS basejails.
-
-        When a jail is a NullFS basejail, this list represent the corresponding
-        fstab lines that mount the release.
-        """
-        try:
-            self.jail.release
-        except AttributeError:
-            return []
-
-        if self.jail.config["basejail_type"] != "nullfs":
-            return []
-
-        basedirs = libioc.helpers.get_basedir_list(
-            distribution_name=self.host.distribution.name
-        )
-
-        fstab_basejail_lines = []
-        release_root_path = "/".join([
-            self.jail.release.root_dataset.mountpoint,
-            f".zfs/snapshot/{self.jail.release_snapshot.snapshot_name}"
-        ])
-        for basedir in basedirs:
-
-            source = f"{release_root_path}/{basedir}"
-            destination = f"{self.jail.root_dataset.mountpoint}/{basedir}"
-            fstab_basejail_lines.append(FstabBasejailLine({
-                "source": source,
-                "destination": destination,
-                "type": "nullfs",
-                "options": "ro",
-                "freq": 0,
-                "passno": 0,
-                "comment": self.AUTO_COMMENT_IDENTIFIER
-            }))
-
-        return fstab_basejail_lines
 
     def __iter__(self) -> typing.Iterator[typing.Union[
         FstabAutoPlaceholderLine,
@@ -661,7 +604,6 @@ class JailFstab(Fstab):
         The output includes user configured and auto created lines for NullFS
         basejails. The previous position of auto-created entries is preserved.
         """
-        basejail_lines_added = False
         output: typing.List[
             typing.Union[
                 FstabAutoPlaceholderLine,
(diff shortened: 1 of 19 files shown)


libioc/Jail.py
```diff
@@ -1832,39 +1660,69 @@ def get_hook_script_path(self, hook_name: str) -> str:
...
+            yield from self._stop_vimage_network(jid, event_scope=event_scope)
+        else:
+            # ToDo: clarify why this step is wanted
+            return
+            yield from self._stop_non_vimage_network(
+                force=force,
+                event_scope=event_scope
+            )
+
+    def _stop_non_vimage_network(
```
igalic: on the one hand i'm thinking, there has to be a better name for this
but on the other hand, this kind of shows our preference ;)

libioc/Network.py
```diff
@@ -211,68 +220,50 @@ def epair_id(self) -> int:
...
-        commands.append(
-            f"export {variable_name_b}="
-            f"$(echo ${variable_name_a} | sed 's/.$/b/')"
         )
+        nic_b_name = nic_a.name[:-1] + "b"
 
-        epair_a = libioc.NetworkInterface.QueuingNetworkInterface(
-            name=None,
+        libioc.NetworkInterface.NetworkInterface(
+            name=nic_a,
```
igalic: can you briefly elaborate what it is that makes the queueing of commands unnecessary now?

libioc/Network.py
```diff
@@ -211,68 +220,50 @@ def epair_id(self) -> int:
...
-
-        commands_created: typing.List[str] = []
-        commands_start: typing.List[str] = []
+        return nic_a, nic_b
 
-        if self.bridge is None:
-            raise libioc.errors.VnetBridgeMissing(logger=self.logger)
+    def __create_vnet_iface(
+        self
+    ) -> typing.Generator['libioc.events.IocEvent', None, None]:
```
igalic: this line could be a wrapper type

libioc/Network.py
```diff
@@ -397,67 +375,42 @@ def __configure_firewall(self, mac_address: str) -> typing.List[str]:
...
-                self.queuing_firewall.add_rule(firewall_rule_number, [
+                self.firewall.add_rule(firewall_rule_number, [
                     "allow", protocol,
                     "from", _address, "to", "any",
                     "layer2",
                     "MAC", "any", mac_address,
-                    "via", f"{self._escaped_nic_name}:$IOC_JID:b",
+                    "via", f"{self._escaped_nic_name}:{self.jail.jid}:b",
                     "out"
-                ], insecure=True)
```
igalic: what does (the removal of) this (parameter) mean?

libioc/helpers.py
```diff
@@ -532,35 +535,27 @@ def exec_passthru(
...
+    data = dict(
+        fstype=fstype,
+        fspath=destination
+    )
+    if source is not None:
+        data["target"] = source
+    for opt in opts:
+        data[opt] = None
+    jiov = libjail.Jiov(data)
+    if libjail.dll.nmount(jiov.pointer, len(jiov), 0) != 0:
```
igalic: i understand that you're basically just using py-jail as libc interface, but it feels dirty 😅
maybe write that in a function's comment

libioc/Config/Jail/File/Fstab.py
```diff
@@ -596,59 +581,19 @@ def parse_lines(
...
-        # skip lines with destinations overlapping self.basejail_lines
-        skip_destinations += list(map(
-            lambda x: str(x['destination']),
-            self.basejail_lines
-        ))
+        BasejailStorage = libioc.Storage.Basejail.BasejailStorage
+        if isinstance(self.jail.storage_backend, BasejailStorage) is True:
+            skip_destinations += list(map(
+                lambda x: str(x[1]),
+                self.jail.storage_backend.basejail_mounts
```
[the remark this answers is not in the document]
gronke: .. in this case a conditional would be required. But such a case already exists: https://github.com/bsdci/libioc/blob/e903f88d4eeb6211771ae62f819024eba8ab66ff/libioc/helpers.py#L611-L628

libioc/Jail.py
```diff
@@ -313,7 +313,6 @@ def __init__(
         self.logger = libioc.helpers_object.init_logger(self, logger)
         self.zfs = libioc.helpers_object.init_zfs(self, zfs)
         self.host = libioc.helpers_object.init_host(self, host)
-        self._relative_hook_script_dir = "/.iocage"
```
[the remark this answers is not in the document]
gronke: They're gone!

The reason they existed was that everything was done in the poststart hook. Due to more control over the jail start process, those have been replaced with individual events that have deprecated all the shell scripts.

libioc/Jail.py
```diff
@@ -545,12 +480,51 @@ def _stop_failed_jail(
             for jail_to_stop in jails_to_stop:
                 yield from jail_to_stop.stop(
                     force=True,
-                    event_scope=jailLaunchEvent.scope
+                    event_scope=jailAttachEvent.scope
+                )
+        jailAttachEvent.add_rollback_step(_stop_failed_jail)
+        jiov = libjail.Jiov(self._launch_params)
+        jid = libjail.dll.jail_set(jiov.pointer, len(jiov), 1)
```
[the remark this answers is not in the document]
gronke: These interfaces come from https://github.com/gronke/py-jail - if you have suggestions for the interfaces, please fell invited to create an Issue on the repo.

libioc/Jail.py
```diff
@@ -429,114 +428,50 @@ def start(
...
 
+        # Start Dependant Jails
         dependant_jails_started: typing.List[JailGenerator] = []
         if start_dependant_jails is True:
             dependant_jails_seen.append(self)
             DependantsStartEvent = libioc.events.JailDependantsStart
             for event in self._start_dependant_jails(
                 self.config["depends"],
-                event_scope=event_scope,
+                event_scope=jailStartEvent.scope,
```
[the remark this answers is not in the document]
gronke: It's not new. By passing an event_scope, the parent can be defined. Visually this causes the event printer of the CLI to nest them and internally they belong to a group that fails together.

Btw. I'm thinking of a way to render the stacks of events that the generator functions yield, so that they can be put in the Wiki or Handbook.

libioc/Jail.py
```diff
@@ -563,41 +537,95 @@ def _stop_failed_jail(
...
-                    yield from self.__stop_network(jid, False, scope)
-                yield from self.__teardown_mounts(False, scope)
-                yield from self.__clear_resource_limits(False, scope)
-
-            if returncode != 0:
-                raise libioc.errors.JailLaunchFailed(
-                    jail=self,
+            yield from self.__teardown_mounts(False, jailStartEvent.scope)
+            yield from self.__clear_resource_limits(False, scope)
+            yield from self.__unmount_devfs(jailStartEvent.scope)
```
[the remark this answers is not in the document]
gronke: Partially yes, but single-command execution and persistent jails require different hooks to be triggered. It is more error prone to repeat the other lines (because it can be easily forgotten on changes), but on the other hand reduces call stack complexity, which I find a good trade in this case.

libioc/Jail.py
gronke: What should it be called? Regular Network? 🤷‍♂️

libioc/Network.py
gronke: Queuing* was a hack around our inability to execute right away. The Queuing wrapper class collected all calls to self._exec and presented the list of commands at a later time. From those lists, the jail hooks have been previously generated. They have been made from minute 1 to get deprecated in a moment like this.

libioc/Network.py
gronke: How do you mean?

libioc/Network.py
gronke: We no longer need a shell (that resolves env variables). We know the jails JID at the time of running this code now, so that there is simply no need for it.

libioc/helpers.py
gronke: > i understand that you're basically just using py-jail as libc interface, but it feels dirty 😅

Yes, it does. The reason to re-use the DLL here is that it was already loaded by py-jail (https://github.com/gronke/py-jail/blob/master/jail/libc.py). It would have been possible to use the dll from py-freebsd_sysctl (https://github.com/gronke/py-freebsd_sysctl/blob/master/freebsd_sysctl/libc.py), but in either case the result is the same.

Although this are just 3 LOC, I didn't want to repeat them a third time in libioc. It's a public interface of either py-jail or py-freebsd_sysctl, so that it should not make a difference - even when calling nmount, which is totally unrelated to both (jail and sysctl).

More libc wrappers are coming, but I do not see the big picture how they all will belong together yet. But this is a process happening outside of this repository - but at least we can detect breakage when updating the dependencies and respond accordingly.

libioc/Jail.py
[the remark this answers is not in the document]
igalic: that means that we have to update the provisioning plugins

libioc/Jail.py
igalic: well, it's what other containers and virtualuzations call a host-network, but that, from inside this code, can be confusing

libioc/Jail.py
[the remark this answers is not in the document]
gronke: Great catch! Yes, we do need to update them.

PR #702, opened 2019-04-05
fix usage of multiple resource limits

### LINKED ISSUE #699: Jail with multiple resource limits cannot be started
Looks like that df0ad67df1f63450c01f3bccffdad73c5f569fa1 broke `ioc create`. I get error like this

```
[-] JailStart@ioc/ancient-step-0E4WGX: ...
  [+] JailDependantsStart@ioc/ancient-step-0E4WGX: No dependant jails [0.0s]
  [+] JailResolverConfig@ioc/ancient-step-0E4WGX: OK [0.001s]
  [-] JailResourceLimitAction@ioc/ancient-step-0E4WGX: ...
  [-] JailResourceLimitAction@ioc/ancient-step-0E4WGX: FAILED [0.005s]
l
    rctl: malformed rule '-a': missing subject
    rctl: malformed rule '-a': missing subject
Command exited with 2
```
So looks like something around rctl inside Jail.py It's still broken on latest master

I'll add this as well, that creating simple jail without any resource limits IP address is not set when jail is started. Not sure if config changed but last working version is e903f88d4eeb6211771ae62f819024eba8ab66ff

gronke:
fixes #699

- call /usr/bin/rctl for each resource limit individually

### Future Work
- Replace rctl command forks with calls to sys/rctl.h

### DIFF
--- a/libioc/Jail.py
+++ b/libioc/Jail.py
@@ -1830,7 +1830,7 @@ class JailGenerator(JailResource):
 
     @property
     def __resource_limits_enabled(self) -> bool:
-        return (self.config["rlimits"] is False) is False
+        return (self.config["rlimits"] is True)
 
     def __clear_resource_limits(
         self,
--- a/libioc/Jail.py
+++ b/libioc/Jail.py
@@ -521,9 +521,9 @@ class JailGenerator(JailResource):
 
         # Setup Network
         if self.config["vnet"]:
-            yield from self._start_vimage_network(jailStartEvent.scope)
-            yield from self._configure_localhost_commands(jailStartEvent.scope)
-            yield from self._configure_routes_commands(jailStartEvent.scope)
+            yield from self._start_vimage_network()
+            yield from self._configure_localhost_commands()
+            yield from self._configure_routes_commands()
             if self.host.ipfw_enabled is True:
                 self.logger.verbose(
                     f"Disabling IPFW in the jail {self.full_name}"
@@ -532,9 +532,7 @@ class JailGenerator(JailResource):
 
         # Attach shared ZFS datasets
         if self.config["jail_zfs"] is True:
-            yield from self._zfs_share_storage.mount_zfs_shares(
-                event_scope=jailStartEvent.scope
-            )
+            yield from self._zfs_share_storage.mount_zfs_shares()
 
         if quick is False:
             unknown_config_parameters = list(
@@ -1801,12 +1799,11 @@ class JailGenerator(JailResource):
         )
         yield event.begin()
 
-        args: typing.List[str] = []
-
         if self.config['rlimits'] is False:
             yield event.skip("disabled")
             return
 
+        skipped = True
         for key in libioc.Config.Jail.Properties.ResourceLimit.properties:
             try:
                 rlimit_prop = self.config[key]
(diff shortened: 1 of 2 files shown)


libioc/Jail.py
```diff
@@ -1802,6 +1802,7 @@ def _apply_resource_limits(
         yield event.begin()
 
         args: typing.List[str] = []
+        skipped = True
```
igalic: so, this mea,…

libioc/Jail.py
```diff
@@ -1816,22 +1817,21 @@ def _apply_resource_limits(
...
 
             rule = f"jail:{self.identifier}:{key}:{rlimit_prop.limit_string}"
-            args.append("-a")
-            args.append(rule)
+            try:
+                libioc.helpers.exec(
+                    ["/usr/bin/rctl", "-a", rule],
+                    logger=self.logger
+                )
+                skipped = False
```
igalic: …we got all the way here without failure.

libioc/Jail.py
```diff
@@ -1816,22 +1817,21 @@ def _apply_resource_limits(
...
+            except Exception:
+                yield event.fail()
+                raise
 
-        if len(args) == 0:
+        if skipped is True:
```
igalic: how would this be true?

libioc/Jail.py
```diff
@@ -1802,6 +1802,7 @@ def _apply_resource_limits(
         yield event.begin()
 
         args: typing.List[str] = []
+        skipped = True
 
         if self.config['rlimits'] is False:
             yield event.skip("disabled")
```
igalic: given the `return` below here? (git cuts it off, you'll need to unfold to see it)

libioc/Jail.py
gronke: When no resource limit was applied, this value stays True, so that the event can be properly flagged as skipped.

igalic: are you sure we didn't make a mistake of some sort in that case?

libioc/Jail.py
```diff
@@ -1845,22 +1845,23 @@ def __clear_resource_limits(
...
-            ignore_error=True
+                f"jail:{self.identifier}",
+                "2>&1"
+            ]),
+            ignore_error=True,
+            shell=True
         )
 
-        if returncode > 0:
+        if (returncode > 0) and ("No such process" not in stdout):
```
igalic: wouldn't the string you're looking for be found in `stderr`, which you join (`2>&1`) into `stdout` (which we don't care about at all…)

gronke: Yes, but stderr handling requires a second thread to observe stderr. It is much easier just to merge the two and look at the combined results.

libioc/Jail.py
gronke: Notifying about a skipped event should be the last step of an IocEvent generator function.

gronke:
Unfortunately Resource Limits (`kern.features.rctl=1`) are not enabled on Cirrus Ci and we have to skip ResourceLimit tests. If there are issues with resource limits, offline tests still cause failure and would be detected.

libioc/Jail.py
gronke: No, because Resource Limits are treated as skipped until at least one was applied successfully. Anyhow the following `return` can be removed, which is done in the latest commit.

PR #703, opened 2019-04-06
always open fstab files with UTF-8 encoding

gronke:
- allow opening fstab files with UTF-8 independent of the hosts default charset


igalic:
i thought we already had that for all files…

PR #704, opened 2019-04-06
Release v0.8.0

gronke:
### Breaking
- events during start and fork_exec have changed and are fine grained now

### Enhancements
- increased Jail start performance due to usage of [py-jail](https://github.com/gronke/py-jail)
- improve handling of invalid IP address configuration
- strict JailConfig property validation, but permissive reading of existing invalid config (with warnings)

### Bugs
- fix support for multiple resource limits
- <VNET_NIC>_mac is a known JailConfig property
- install py36-ucl to fix issues importing libucl.so.6
- always open fstab files with UTF-8 encoding
- properly setup and teardown non-VNET network
- support multiple IPv4 and IPv6 addresses when using non-VNET network


igalic:
every file is read as UTF-8 (maybe)

PR #716, opened 2019-04-24
support output of firewall rules

himrock922:
Hi all.

A firewall rules add or delete via libioc, that result wants output.
To implement it, I did commit of Firewall.rules method.
It is very simply, Because, I want review(Also, it's need or unnecessary)
Regards,

Make sure to follow and check these boxes before submitting a PR! Thank you.

- [x] Explain the feature
- [x] Read [CONTRIBUTING.md](https://github.com/bsdci/libioc/blob/master/CONTRIBUTING.md)

### DIFF
--- a/libioc/Firewall.py
+++ b/libioc/Firewall.py
@@ -82,6 +82,16 @@ class Firewall:
             logger=self.logger
         )
 
+    def rules(
+        self
+    ) -> str:
+        """Output lists of firewall rule."""
+        command = [
+            self.IPFW_COMMAND,
+            "-a", "list"
+        ]
+        self._exec(command)
+
     def delete_rule(
         self,
         rule_number: typing.Union[int, str],
@@ -130,7 +140,7 @@ class Firewall:
         ignore_error: bool=False
     ) -> None:
         try:
-            libioc.helpers.exec(command, ignore_error=ignore_error)
+            return libioc.helpers.exec(command, ignore_error=ignore_error)
         except libioc.errors.CommandFailure:
             raise libioc.errors.FirewallCommandFailure(
                 logger=self.logger
--- a/libioc/Firewall.py
+++ b/libioc/Firewall.py
@@ -90,7 +90,7 @@ class Firewall:
             self.IPFW_COMMAND,
             "-a", "list"
         ]
-        self._exec(command)
+        return self._exec_list(command)
 
     def delete_rule(
         self,
@@ -134,13 +134,25 @@ class Firewall:
         _rule_number = int(rule_number)
         return str(_rule_number + self.IPFW_RULE_OFFSET)
 
+    def _exec_list(
+        self,
+        command: typing.List[str],
+        ignore_error: bool=False
+    ) -> str:
+        try:
+            return str(libioc.helpers.exec(command, ignore_error=ignore_error))
+        except libioc.errors.CommandFailure:
+            raise libioc.errors.FirewallCommandFailure(
+                logger=self.logger
+            )
+
     def _exec(
         self,
         command: typing.List[str],
         ignore_error: bool=False
     ) -> None:
         try:
-            return libioc.helpers.exec(command, ignore_error=ignore_error)
+            libioc.helpers.exec(command, ignore_error=ignore_error)
         except libioc.errors.CommandFailure:
             raise libioc.errors.FirewallCommandFailure(
                 logger=self.logger
--- a/libioc/Firewall.py
+++ b/libioc/Firewall.py
@@ -90,7 +90,7 @@ class Firewall:
             self.IPFW_COMMAND,
             "-a", "list"
         ]
-        return self._exec_list(command)
+        return str(self._exec(command))
 
     def delete_rule(
         self,
@@ -134,25 +134,13 @@ class Firewall:
         _rule_number = int(rule_number)
         return str(_rule_number + self.IPFW_RULE_OFFSET)
 
-    def _exec_list(
+    def _exec(
         self,
         command: typing.List[str],
         ignore_error: bool=False
     ) -> str:
         try:
             return str(libioc.helpers.exec(command, ignore_error=ignore_error))
-        except libioc.errors.CommandFailure:
-            raise libioc.errors.FirewallCommandFailure(
-                logger=self.logger
-            )
-
-    def _exec(
-        self,
-        command: typing.List[str],
-        ignore_error: bool=False
-    ) -> None:
-        try:
-            libioc.helpers.exec(command, ignore_error=ignore_error)
         except libioc.errors.CommandFailure:
             raise libioc.errors.FirewallCommandFailure(
                 logger=self.logger


libioc/Firewall.py
```diff
@@ -82,6 +82,16 @@ def ensure_firewall_enabled(self) -> None:
             logger=self.logger
         )
 
+    def rules(
+        self
+    ) -> str:
+        """Output lists of firewall rule."""
+        command = [
+            self.IPFW_COMMAND,
+            "-a", "list"
```
gronke: This command generates a generic list of firewall rules on a host, while the Firewall module should focus on rules that are managed by libioc. Those rules begin at a certain offset - finding the upper boundary is a tougher question.

gronke:
@himrock922 in case you want to get a list of all firewall rules active, another way to get them is to filter the list of all jails for running jails that use Secure VNET. (There are no user managed firewall rules possible yet, so ipfw is only used when Secure VNET (e.g. `vnet0::bridge0`) is enabled.

At the moment, this appears to be the quickest way to get all running jails with Secure VNET enabled.

```python3
[j for j in libioc.Jails("running=True") if any([n._is_secure_vnet_bridge for n in j.networks])]
```

Unfortunately the Firewall class is a minimal write-only abstraction of IPFW. This PR brings us a large step forward. 👍

libioc/Firewall.py
```diff
@@ -124,6 +134,18 @@ def _offset_rule_number(
         _rule_number = int(rule_number)
         return str(_rule_number + self.IPFW_RULE_OFFSET)
 
+    def _exec_list(
+        self,
+        command: typing.List[str],
+        ignore_error: bool=False
+    ) -> str:
+        try:
+            return str(libioc.helpers.exec(command, ignore_error=ignore_error))
```
gronke: Please apply this change directly to the `_exec` method and change its return type to `str`. It will not interfere with other methods using _exec, as they will all ignore the "returned" `None`.

himrock922:
@gronke 

```
 Getting a generic list of ipfw rules on a host, seems to be a bit out of scope for libioc, but I am convinced that the ability to verify if the expected rules are set, can be helpful.

---
Unfortunately the Firewall class is a minimal write-only abstraction of IPFW. This PR brings us a large step forward.
```

Thanks, I thinks so it too.
Rather, you might as well detach firewall function from libioc?
It is firewall operations only python library.

libioc/Firewall.py
himrock922: right.
Honestly, I think shouldn't setting firewall for host via libioc too.
But, I think correct that setting(or confirm) firewall related host and container on same library. Because It is layer, If library doesn't deal both, that is irrelevant.

gronke:
> Rather, you might as well detach firewall function from libioc?

That's a great idea! I'll investigate our options and invite you to a repository.

himrock922:
> That's a great idea! I'll investigate our options and invite you to a repository.

@gronke 
Thanks!
I'm waiting for your response.

gronke:
> > That's a great idea! I'll investigate our options and invite you to a repository.
> 
> @gronke
> Thanks!
> I'm waiting for your response.

Unfortunately there is no libc interface for ipfw. I've looked into the sbin/ipfw implementation of FreeBSD but did not come up with a proper idea to implement this in native Python 3: https://github.com/freebsd/freebsd/blob/50b3d8bf81088f77cb3a48d9d9f7f368ecfb42b5/sbin/ipfw/ipfw2.c#L2685-L2742

As so often, @fabianfreyer could have the idea 😍

I would not like to implement a wrapper that parses sbin/ipfw stdout, because everyone could already do that on their own today. Why making it more complex then? But with programmatic access to IPFW rules, having a library would make so much sense.

himrock922:
Hi @gronke 

> Unfortunately there is no libc interface for ipfw. I've looked into the sbin/ipfw implementation of FreeBSD but did not come up with a proper idea to implement this in native Python 3: https://github.com/freebsd/freebsd/blob/50b3d8bf81088f77cb3a48d9d9f7f368ecfb42b5/sbin/ipfw/ipfw2.c#L2685-L2742

Thank you for researching and I'm sorry...

> I would not like to implement a wrapper that parses sbin/ipfw stdout, because everyone could already do that on their own today. Why making it more complex then?

Right. I'm think it too.

But, If container tools development, I'm think that need firewall function(or, nat)  flatly handling too.
(Let's put aside it, the libioc need or unnecessarily.)

> But with programmatic access to IPFW rules, having a library would make so much sense.

Right. I'm think it too. this reason ↑.

Once, I'd develop it, the library for firewall function handling.

fabianfreyer:
One question that comes to my mind is: why ipfw specifically over the other options for firewalling?

himrock922:
Hi, @fabianfreyer 

> One question that comes to my mind is: why ipfw specifically over the other options for firewalling?
> Why not, e.g. [pf(4)](https://www.freebsd.org/cgi/man.cgi?query=pf&sektion=4&manpath=freebsd-release-ports?)? Afaik, [npf(7)](https://netbsd.gw.com/cgi-bin/man-cgi?npf+7+NetBSD-current) [is being ported to FreeBSD](https://github.com/gvnn3/npf/tree/gnn/freebsd) too, and has [libnpf(3)](https://netbsd.gw.com/cgi-bin/man-cgi?libnpf+3+NetBSD-current). [ipf(4)](https://www.freebsd.org/cgi/man.cgi?query=ipf&sektion=4&apropos=0&manpath=FreeBSD+12.0-RELEASE+and+Ports) also has a programmatic interface. The ipfw programmatic interface seems to be undocumented at this point, so I'm not sure how stable it would be to write bindings for it directly.

Honestly, Because the libioc uses ipfw, I did only used ipfw too.

> For an example of a [pf(4)](https://www.freebsd.org/cgi/man.cgi?query=pf&sektion=4&manpath=freebsd-release-ports?) consumer, take a look at [filter.c](https://github.com/freebsd/freebsd/blob/1d6e4247415d264485ee94b59fdbc12e0c566fd0/contrib/pf/ftp-proxy/filter.c) in [ftp-proxy(8)](https://www.freebsd.org/cgi/man.cgi?query=ftp-proxy&sektion=8&n=1).

Thanks! I'm read it.


> Finally, you could also use [divert(4)](https://www.freebsd.org/cgi/man.cgi?query=divert&sektion=4&apropos=0&manpath=FreeBSD+12.0-RELEASE+and+Ports) to write your own filter (though ~I'm pretty sure that's not a good idea~ this _may_ be a good idea with [npf(7)](https://netbsd.gw.com/cgi-bin/man-cgi?npf+7+NetBSD-current)).

Yes, Actually, I'm setting for firewall via ipfw and divert.
But, I'm want setting it via web application.
Actually I'm development it.

https://github.com/himrock922/jaisting

fabianfreyer:
> This is misunderstanding.

ah, no, I was just dreaming ;)

gronke:
> One question that comes to my mind is: why ipfw specifically over the other options for firewalling?

Because we need layer 2 filtering on bridges to mitigate ARP Spoofing from jails that use the Secure VNET feature (`host bridge <> epair <> filter bridge <> epair <> jail`).

himrock922:
Hi @gronke 
> > One question that comes to my mind is: why ipfw specifically over the other options for firewalling?
> 
> Because we need layer 2 filtering on bridges to mitigate ARP Spoofing from jails that use the Secure VNET feature (`host bridge <> epair <> filter bridge <> epair <> jail`).

Is this answer that  ipfw is better  compared to other firewall management tools?

Can't blocking other tools ARP spoofing...?

fabianfreyer:
@himrock922 it means that ipfw is an OSI layer 2 firewall, while the others are OSI layer 3. Libioc puts jails on the same layer2 domain, therefore uses ipfw. If it would not do that (but instead e.g. route and/or redirect), a layer 3 firewall would be fine as well.

So no, in general it is not "better", just a better tool for this specific job given the constraints.

gronke:
Manipulation of existing host network devices is not performed by libioc. A VNET jail can be connected to a bridge, but we do not setup virtual routing for it (yet).

The network configuration was pretty much inherited from iocage. For advanced network configuration, such as NAT or local routing (instead of bridging), I'd like to overthink this concept and watch out for suitable standards, such as [opencontainers](https://github.com/opencontainers/runc/blob/master/libcontainer/configs/network.go).

himrock922:
> The network configuration was pretty much inherited from iocage. For advanced network configuration, such as NAT or local routing (instead of bridging), I'd like to overthink this concept and watch out for suitable standards, such as [opencontainers](https://github.com/opencontainers/runc/blob/master/libcontainer/configs/network.go).

Right.To configuration advanced network for container management,
I agree for wait there will establish standards or draft.

Temporary, I'll process development to establish my container hosting system.

himrock922:
Hi @gronke.
https://github.com/himrock922/py-ipfw

I had develop IPFW wrapper for Python.
I had only implement ipfw -a list. Therefore, the libioc doesn't need for firewall results.
I do continue that development of IPFW wrapper.

You can close this PR.
Thanks.

gronke:
Sounds great, @himrock922.

In the near future I'll look into routing (instead of layer2 filtering) as solution to move firewall tasks to PF. That's probably the right time to make sure firewall rules are configured through such a wrapper.

himrock922:
Hi All.
https://github.com/himrock922/py-ipfw

py-ipfw has supported function add and delete for ipfw by this development.
Therefore, this PR close.

Thanks.

igalic:
*starts watching*

gronke:
> starts watching

same here :) I'd be happy to replace the IPFW wrapper with a real library that does proper parsing of firewall rules. I would be even more happy to replace IPFW layer 2 filtering with PF, but that's a totally different task.

PR #727, opened 2019-05-08
Mount fdescfs on jail start when mount_fdescfs is set

gronke:
# Security
- mount /dev and /dev/fd from within the jail
- apply security checks before unmounting system mountpoints from a jail

# Bugs
- fix mounting /dev when mount_devfs is enabled
- fix mounting /dev/fd when mount_fdescfs is enabled
- include sysvshm, sysvsem and sysvmsg jail params on start

---

`mount_descfs` is a /usr/sbin/jail argument, not a Jail parameter. Therefore it is not handled by py-jail and must be manually handled during jail start.

### DIFF
--- a/libioc/Config/Jail/BaseConfig.py
+++ b/libioc/Config/Jail/BaseConfig.py
@@ -621,7 +621,7 @@ class BaseConfig(dict):
     def unknown_config_parameters(self) -> typing.Iterator[str]:
         """Yield unknown config parameters already stored in the config."""
         for key in self.data.keys():
-            if self._is_known_property(key, explicit=True) is False:
+            if self.is_known_property(key, explicit=True) is False:
                 yield key
 
     def __delitem__(self, key: str) -> None:
@@ -636,7 +636,7 @@ class BaseConfig(dict):
         explicit: bool=True
     ) -> None:
         """Set a configuration value."""
-        if self._is_known_property(key, explicit=explicit) is False:
+        if self.is_known_property(key, explicit=explicit) is False:
             if "jail" in dir(self):
                 _jail = self.jail  # noqa: T484
             else:
@@ -865,7 +865,7 @@ class BaseConfig(dict):
         """Return whether the given key belongs to a custom user property."""
         return (key == "user") or (key.startswith("user.") is True)
 
-    def _is_known_property(self, key: str, explicit: bool) -> bool:
+    def is_known_property(self, key: str, explicit: bool) -> bool:
         """Return True when the key is a known config property."""
         if self._is_known_jail_param(key):
             return True
@@ -898,7 +898,7 @@ class BaseConfig(dict):
         key: str,
         explicit: bool=True
     ) -> None:
-        if self._is_known_property(key, explicit=explicit) is False:
+        if self.is_known_property(key, explicit=explicit) is False:
             raise libioc.errors.UnknownConfigProperty(
                 key=key,
                 logger=self.logger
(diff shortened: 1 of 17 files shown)


libioc/Config/Jail/BaseConfig.py
```diff
@@ -343,6 +343,35 @@ def _set_exec_clean(self, value: typing.Union[str, int, bool]) -> None:
...
+        else:
+            try:
+                str_value = str(int(value))
+            except ValueError:
+                error_reason = "invalid input type (expected Integer or String)"
+
+            if (str_value != "0") and (str_value != "1"):
+                error_reason = f"Boolean input expected, but got {str_value}"
+
+            enabled = (str_value == "1")
```
igalic: didn't we have a helper function for this whole mess here?

libioc/Jail.py
```diff
@@ -630,27 +633,40 @@ def __mount_devfs(
 
         yield event.end()
 
-    def __unmount_devfs(
+    def __mount_fdescfs(
         self,
```
igalic: is this just git diff weirdness, or why did `__umount_devfs` go?

gronke: __umount_devfs was unused because Storage.teardown already unmounts `/dev/fd` and `/dev`.

requirements.txt
```diff
@@ -1,3 +1,3 @@
 gitpython
-freebsd_sysctl==0.0.5
+freebsd_sysctl==0.0.6
```
igalic: there's no official release https://github.com/gronke/py-freebsd_sysctl/releases

libioc/Config/Jail/Globals.py
```diff
@@ -88,7 +88,7 @@
     "exec_timeout": "600",
     "stop_timeout": "30",
     "mount_procfs": "0",
-    "mount_devfs": "1",
+    "mount_devfs": "0",
```
igalic: oh?

can you explain this change?

requirements.txt
gronke: Release tag is published. Thanks for the heads up!

libioc/Config/Jail/Globals.py
gronke: The default property `allow_mount_devfs` on iocage was `0`, while `mount_devfs` was `1`. Combined, the jail is not allowed to mount it's own devfs.

```console
$ ioc create myjail allow_mount_devfs=no mount_devfs=yes
myjail successfully created from 12.0-RELEASE!
$ ioc start myjail
[-] JailStart@myjail: ...
  [+] JailDependantsStart@myjail: No dependant jails [0.0s]
  [+] JailResolverConfig@myjail: OK [0.001s]
  [+] JailResourceLimitAction@myjail: disabled [0.001s]
  [+] JailHookPrestart@myjail: SKIPPED [0.0s]
  [+] BasejailStorageConfig@myjail: OK [0.024s]
  [+] MountFstab@myjail: OK [0.0s]
  [+] JailAttach@myjail: OK [0.021s]
  [+] JailHookCreated@myjail: SKIPPED [0.0s]
  [-] MountDevFS@myjail: FAILED [0.047s]
Failed to mount /dev: mount: devfs: Operation not permitted
```

Because of that, either of the defaults needed be changed. Choosing the lower permissions as default, appears to be the right choice here. Otherwise the default `enforce_statfs` would be `1`, not `2` as defined in the Globals.

PR #728, opened 2019-05-10
use jail devfs_ruleset instead of always 4

gronke:
Uses the (auto-)configured devfs_ruleset jail config property instead of always assuming 4 as DevFS ruleset.


igalic:
can you explain *how* this fixes the bug?

gronke:
Before the ruleset was hardcoded. With this changes, the ruleset is taken from kwargs of the method.

PR #729, opened 2019-05-10
fix bootstrap of pkg from archive

gronke:
- bootstrap the newest pkg archive found in the mirror cache

The local mirror cache was mistakenly searched for the first file that begins with `pkg` and ends with `.txz`. Instead the list of matches is sorted in descending order and the first item (the highest version) is returned. Also the package must start with `pkg-` (dash added).

### DIFF
--- a/libioc/Jail.py
+++ b/libioc/Jail.py
@@ -872,7 +872,7 @@ class JailGenerator(JailResource):
         self.config.clone(original_config.data)
 
         try:
-            fork_exec_events = JailGenerator.start(
+            yield from JailGenerator.start(
                 self,
                 single_command=command,
                 passthru=passthru,
@@ -881,8 +881,6 @@ class JailGenerator(JailResource):
                 start_dependant_jails=start_dependant_jails,
                 env=env
             )
-            for event in fork_exec_events:
-                yield event
         finally:
             self.config = original_config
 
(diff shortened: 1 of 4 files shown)


libioc/Pkg.py
```diff
@@ -67,6 +67,7 @@ def fetch(
         event_scope: typing.Optional['libioc.events.Scope']=None
     ) -> typing.Generator[libioc.events.IocEvent, None, None]:
         """Fetch a bunch of packages to the local mirror."""
+        return
```
igalic: is this a good idea?

gronke: No, it was not! We used this to reduce the delay while debugging.

libioc/Jail.py
```diff
@@ -589,50 +613,35 @@ def _stop_failed_jail(
...
+                    event_scope=jailStartEvent.scope,
+                    script=str(single_command),
+                    passthru=passthru,
+                    env=env
+                )
+        except Exception as e:
+            yield jailStartEvent.fail(e)
+            raise e
 
+        _stop_jails()
```
igalic: we don't catch the events from this function?

gronke: Thanks for spotting the issue. It was patched with the latest commits.

PR #731, opened 2019-05-11
Release v0.8.1

gronke:
### Security
- devfs is mounted with proper devfs_ruleset

### Bugs
- fix package installation with Pkg
- mount fdescfs when mount_fdescfs is enabled
- sendmail is disabled via periodic.conf

### Enhancements
- automatically set enforce_statfs=1 when allowing mounts in jails
- improved error handling of Jail.fork_exec


igalic:
next time, please make sure this PR message is also be the commit's message, otherwise it's a bit confusing when looking thru the git log.

PR #734, opened 2019-05-13
fix calculation of devfs_rules (for DHCP)

gronke:
- mount /dev with calculated devfs_rule rather than the config value

### DIFF
--- a/tests/test_VNET.py
+++ b/tests/test_VNET.py
@@ -142,3 +142,25 @@ class TestVNET(object):
             f"vnet7:{existing_jail.jid}"
         ]).decode("utf-8")
         assert mac_a in stdout
+
+    def test_devfs_ruleset_for_dhcp(
+        self,
+        existing_jail: 'libioc.Jail.Jail',
+        bridge_interface: str
+    ) -> None:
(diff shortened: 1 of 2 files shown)


igalic:
how hard do you think would it be to add a test for dhcp?

Can we improve the docs for these attributes?

tests/test_VNET.py
```diff
@@ -142,3 +142,25 @@ def test_vnet_mac_address_can_be_configured(
...
+
+        assert existing_jail.devfs_ruleset > 4
+        stdout = subprocess.check_output([
+            "/sbin/devfs",
+            "rule",
+            "-s",
+            str(existing_jail.devfs_ruleset),
+            "show"
+        ]).decode("utf-8")
+        assert "bpf" in stdout
```
igalic: could you also check that the process dhclient is there?

gronke: The jail is not started during this test because there is no DHCP Server that can handle the request. The latest commit starts the jail and checks for the availability of the BPF device.

Indeed, it would be interesting to know if the jail can query an IP address with dhclient, but with this test we have now ensured that it's technically possible.

PR #738, opened 2019-06-04
Fix missing IOC_* env variables in Jail hooks

### LINKED ISSUE #342: Hook scripts missing IOCAGE_ env variables

gronke:
closes #342

Env variables from the `JailGenerator.env` did no longer show up in Jail Hook env variables.

- Feed correctly merged JailGenerator.env to Jail hooks
- Test that IOC_ env variables exist in every JailHook event that has finished

This regression was introduced with https://github.com/bsdci/libioc/commit/2e13eb06cf5b459b23124028bb3d324d5e9e306c

### DIFF
--- a/libioc/Jail.py
+++ b/libioc/Jail.py
@@ -899,7 +899,7 @@ class JailGenerator(JailResource):
             if key not in _env_keys:
                 _env[key] = global_env[key]
 
-        return env
+        return _env
 
     def __run_hook(
         self,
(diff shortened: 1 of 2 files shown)


gronke:
@urosgruber finally a unit test for this case. Thanks for your patience!

libioc/Jail.py
```diff
@@ -899,7 +899,7 @@ def __merge_env(
             if key not in _env_keys:
                 _env[key] = global_env[key]
 
-        return env
+        return _env
```
igalic: ooooooof, didn't you bi this would be it

PR #739, opened 2019-07-01
Test and fix UFS mounts from fstab into a jail

gronke:
- Test ability to mount UFS volumes into a jail

FYI @FabianFreyer

### DIFF
--- a/tests/test_Fstab.py
+++ b/tests/test_Fstab.py
@@ -22,7 +22,13 @@
 # IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
 # POSSIBILITY OF SUCH DAMAGE.
 """Unit tests for Fstab module."""
+import pytest
 import tempfile
+import random
+import sys
+import subprocess
+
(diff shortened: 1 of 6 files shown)


tests/test_Fstab.py
```diff
@@ -87,4 +93,33 @@ def test_parses_auto_lines(
...
+		name = f"{root_dataset.name}/zvol{r}"
+
+		root_dataset.pool.create(
+			dataset_name,
+			fsopts=dict(volsize="16M"),
+			fstype=libzfs.DatasetType.VOLUME
+		)
+
+		dataset = zfs.get_dataset(dataset_name)
+		yield dataset
```
fabianfreyer: should we not wrap this generator with something that calls `newfs`?

gronke: Would you agree that another fixture `zfs_volume_ufs` could do that job?

fabianfreyer: probably a good idea

gronke: done

Issue #483, opened 2018-08-26
Iocage jails with dots fail to start


gronke:
Related to https://github.com/iocage/iocage/issues/588

When an iocage jail has dots in its name the jail command fails starting them.

gronke:
Rather than replacing the characters libiocage should prevent jail names including dots.

Corvan:
But why should dots not be allowed?
We use them in our blueprint-jail names and at least iocage works with them as well
https://github.com/iocage/iocage/pull/935

gronke:
On FreeBSD jails are not allowed to have a dot (`.`) in their name identifier. A workaround would be to replace dots with a different character, but since all other characters are allowed, this would lead to the problem of conflicting names.

iocage/iocage#935 does replace `.` with `_`, but imagine somebody has a jail `foo.bar` and `foo_bar` on the same system.

I agree that having dots in jail names can be useful, but I don't see an elegant way to solve this without adding a layer of abstraction, such as for using hashes of the ioc jail name as identifier instead. This would come with the disadvantage of making it more complicated to use standard FreeBSD tooling to interface with jails.

We could decide to sacrifice another character that is no longer allowed in jail names to replace the dot. After some playing around a bit, it seems that a NULL byte does a decent job:

```console
# jail -c persist name="foo.bar" path="/rescue"
jail: jail "foo" not found
# jail -c persist name="foo\x00bar" path="/rescue"
# jail -r "foobar"
jail: "foobar" not found
# jail -r "foo\x00bar"
```

@Corvan do you think that is a good idea? It would only appear for internal identification of jails.

urosgruber:
How about `@` character. At least one case I could use this is to differentiate between the versions or jails?

igalic:
why, exactly, do we not allow `.` in jail names?

gronke:
> How about @ character. At least one case I could use this is to differentiate between the versions or jails?

Yeah, we should be more permissive with allowed characters in jails. While dot characters would need to be replaced, other characters could have a negative on the CLI tool. The question is now if we should care or just allow every character, including different white space characters and ANSI control characters.

[...]

PR #741, opened 2019-08-12
enhance resource name validation

gronke:
addresses #483 

- be more strict with jail names
- allow certain special characters in jail names `! ^ - _ ( ) [ ] { } < > , .`

### DIFF
--- a/libioc/errors.py
+++ b/libioc/errors.py
@@ -393,7 +393,8 @@ class InvalidJailName(JailConfigError):
     ) -> None:
         msg = (
             f"Invalid jail name '{name}': "
-            "Names may only contain alphanumeric characters and dash"
+            "Names may only contain alphanumeric characters and may not begin "
+            "with special characters ! ^ - _ ( ) [ ] { } < > ,"
         )
         if invalid_characters is not None:
             msg += ", but got " + str("".join(invalid_characters) + "")
--- a/libioc/errors.py
+++ b/libioc/errors.py
@@ -393,8 +393,9 @@ class InvalidJailName(JailConfigError):
     ) -> None:
         msg = (
             f"Invalid jail name '{name}': "
-            "Names may only contain alphanumeric characters and may not begin "
-            "with special characters ! ^ - _ ( ) [ ] { } < > , ."
+            "Names may not begin or end with special characters, "
+            "but may contain alphanumeric and allowed special characters "
+            "! ^ - _ ( ) [ ] { } < > , ."
         )
         if invalid_characters is not None:
             msg += ", but got " + str("".join(invalid_characters) + "")
--- a/libioc/errors.py
+++ b/libioc/errors.py
@@ -394,7 +394,7 @@ class InvalidJailName(JailConfigError):
         msg = (
             f"Invalid jail name '{name}': "
             "Names may only contain alphanumeric characters and may not begin "
-            "with special characters ! ^ - _ ( ) [ ] { } < > ,"
+            "with special characters ! ^ - _ ( ) [ ] { } < > , ."
         )
         if invalid_characters is not None:
             msg += ", but got " + str("".join(invalid_characters) + "")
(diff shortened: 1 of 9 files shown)


gronke:
### @ Character
@urosgruber the `@` character was not included in the permitted characters, because it is reserved to identify jail snapshots. Allowing the character would complicate the detection of snapshots, which becomes more important when supporting auto-completion.

### Dots in jail names
The dot `.` character has also not been not permitted in jail names, because it is used by FreeBSD jails to express structures of nested jails. The only occasion where this becomes a problem for ioc is when assigning a name parameter to FreeBSD jails (currently `ioc-<JAIL_NAME>`) - without this human-readable identifier /usr/sbin/jls becomes less comfortable to use. This would also break compatibility with older iocage versions. (running jails are identified by this name)

I do understand that it is inviting to name jails after their domain name, but it comes with another disadvantage: the sorting order. Personally I prefer to name jails in reverse FQDN hierarchy (the domain `subdomain.bsd.ci` would turn into `ci-bsd-subdomain`.

The solution implemented in iocage (Python) https://github.com/iocage/iocage/pull/935/files is not acceptable, because it causes name collisions between jails. The two jails `foo.bar` and `foo_bar` would have the same jail identifier, which causes a race condition.

---

It could make sense to sacrifice the `@` character as replacement for dot `.` in the jail identifier. @joachimmathes might suggest change to iocage. This is already a compatibility issue with latest iocgae that would be nice to resolve.

gronke:
Update: dots in jail names are supported now, replacing the character in jail identifiers (Jail param `name`) with `%`, as it appears to be more useful for jail given names. Opinions welcome!

igalic:
> I do understand that it is inviting to name jails after their domain name, but it comes with another disadvantage: the sorting order. Personally I prefer to name jails in reverse FQDN hierarchy (the domain subdomain.bsd.ci would turn into ci-bsd-subdomain.

if you're making such opinionated decisions, it would be good to anchor them in the documentation (and `ioc command help` output)

libioc/errors.py
```diff
@@ -393,7 +393,8 @@ def __init__(
     ) -> None:
         msg = (
             f"Invalid jail name '{name}': "
-            "Names may only contain alphanumeric characters and dash"
+            "Names may only contain alphanumeric characters and may not begin "
+            "with special characters ! ^ - _ ( ) [ ] { } < > , ."
```
igalic: it's unclear from this message, whether or not the name is allowed to *contain* the these characters.

tests/test_Jail.py
```diff
@@ -339,3 +340,55 @@ def test_hooks_can_access_ioc_env_variables(
...
+        assert libioc.helpers.is_valid_name("ioc\t23") is False
+        assert libioc.helpers.is_valid_name("ioc-") is False
+        assert libioc.helpers.is_valid_name("!ioc") is False
+        assert libioc.helpers.is_valid_name("@gronke") is False
+        assert libioc.helpers.is_valid_name("stefan@gronke") is False
+        assert libioc.helpers.is_valid_name("23-ioc-") is False
+        assert libioc.helpers.is_valid_name("white space") is False
+        assert libioc.helpers.is_valid_name("ioc\\jail") is False
+        assert libioc.helpers.is_valid_name("jail{ioc}") is False
+        assert libioc.helpers.is_valid_name("we❤jails") is False
```
igalic: sad times 🙀

libioc/errors.py
gronke: The error message is clarified now. Thanks for spotting it!

gronke:
Jail names may now contain dots (e.g. `jail.example.com`). The implementation deviates from python-iocage, which replaces dots in jail name parameters (used as identifier of managed jails) with dash (`-`). Because dash is allowed in regular jail names, the running state of two jails `my.jail` and `my-jail` would be confused. This implementation replaces dots in FreeBSD jail names with `*` - a character that is not allowed in ioc Jail names.

PR #746, opened 2019-11-01
Fix starting nested jails without devfs_ruleset

gronke:
- allow starting jails without any devfs_ruleset (`None`)


tests/test_Config.py
```diff
@@ -1,4 +1,4 @@
-# Copyright (c) 2017-2019, Stefan Grönke
+2# Copyright (c) 2017-2019, Stefan Grönke
```
igalic: what's up with this?

PR #748, opened 2019-11-02
Correct hostuuid handling

gronke:
### Enhancements
- load Host.uuid from sysctl rather than /etc/hostid
- strictly handle host_hostuuid as uuid.UUID

### Bugs
- fix nested jails attempt to non-existing `/etc/hostid` file


libioc/Config/Jail/BaseConfig.py
```diff
@@ -24,6 +24,7 @@
 # POSSIBILITY OF SUCH DAMAGE.
 """The common base of jail configurations."""
 import typing
+import uuid
```
igalic: is this our patched version?

Issue #750, opened 2020-03-10
Unable to build under 12.1-p2 (python3.7)


karpa13a:
```
root@host01:/var/tmp # uname -a
FreeBSD host01 12.1-RELEASE-p2 FreeBSD 12.1-RELEASE-p2 GENERIC  amd64
root@host01:/var/tmp # git clone https://github.com/bsdci/libioc
Cloning into 'libioc'...
remote: Enumerating objects: 11, done.
remote: Counting objects: 100% (11/11), done.
remote: Compressing objects: 100% (9/9), done.
remote: Total 12395 (delta 2), reused 4 (delta 2), pack-reused 12384
Receiving objects: 100% (12395/12395), 10.36 MiB | 1.24 MiB/s, done.
Resolving deltas: 100% (8746/8746), done.
root@host01:/var/tmp # cd libioc/
root@host01:/var/tmp/libioc # make
pkg install -q -y libucl py36-ucl py36-cython rsync python36 py36-libzfs
pkg: No packages available to install matching 'py36-ucl' have been found in the repositories
*** Error code 70

Stop.
make: stopped in /var/tmp/libioc
root@host01:/var/tmp/libioc # git status
On branch master
Your branch is up to date with 'origin/master'.

nothing to commit, working tree clean

root@host01:/var/tmp/libioc # git log -n 1
commit 1dcdd858ad8c5deda1ec26c5cc1bac90ae718913 (HEAD -> master, origin/master, origin/HEAD)
Author: Stefan Gr<C3><B6>nke <stefan@gronke.net>
Date:   Sat Nov 2 16:17:42 2019 +0000

    check Fstab automount destination directory
root@rbps05-md-host01:/var/tmp/libioc # 

```

PR #751, opened 2020-03-21
Update default python version

gronke:
addresses #750

- Update default Python version to 3.7
- Document installation from Makefile for other Python versions in README

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,8 +4,8 @@ JAIL_IP?=172.16.0
 JAIL_NET?=16
 MYPYPATH = $(shell pwd)/.travis/mypy-stubs
 
-LATEST_PKG_PYTHON_VERSION != pkg info -g 'python3*' | cut -d'-' -f1 | sed 's/^python//' | sort -n | tail -n1 | sed -r 's/^([0-9])([0-9]+)/\1.\2/'
-PYTHON ?= python${LATEST_PKG_PYTHON_VERSION}
+PYTHON_VERSION != if [ "$(PYTHON_VERSION)" != "" ]; then echo $(PYTHON_VERSION); else pkg info -g 'python3*' | cut -d'-' -f1 | sed 's/^python//' | sort -n | tail -n1 | sed -r 's/^([0-9])([0-9]+)/\1.\2/'; fi
+PYTHON ?= python${PYTHON_VERSION}
 pyver= ${PYTHON:S/^python//:S/.//:C/\([0-9]+\)/\1/}
 
 .if $(pyver) < 35
--- a/Makefile
+++ b/Makefile
@@ -22,7 +22,7 @@ install-python-requirements:
 install-python-requirements-dev: install-python-requirements
 	$(PYTHON) -m pip install -Ur requirements-dev.txt
 install-deps:
-	pkg install -q -y libucl py$(pyver)-ucl py$(pyver)-cython rsync python$(pyver) py$(pyver)-libzfs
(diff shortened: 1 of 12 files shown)


Makefile
```diff
@@ -3,8 +3,9 @@ JAIL_NIC?=vtnet0
 JAIL_IP?=172.16.0
 JAIL_NET?=16
 MYPYPATH = $(shell pwd)/.travis/mypy-stubs
-PYTHON ?= python3.6
 
+LATEST_PKG_PYTHON_VERSION != pkg info -g 'python3*' | cut -d'-' -f1 | sed 's/^python//' | sort -n | tail -n1 | sed -r 's/^([0-9])([0-9]+)/\1.\2/'
+PYTHON ?= python${LATEST_PKG_PYTHON_VERSION}
 pyver= ${PYTHON:S/^python//:S/.//:C/\([0-9]+\)/\1/}
```
igalic, line 9: given that we already have `LATEST_PYTHON_VERSION` we can use this directly here

Makefile
```diff
@@ -3,8 +3,9 @@ JAIL_NIC?=vtnet0
 JAIL_IP?=172.16.0
 JAIL_NET?=16
 MYPYPATH = $(shell pwd)/.travis/mypy-stubs
-PYTHON ?= python3.6
 
+LATEST_PKG_PYTHON_VERSION != pkg info -g 'python3*' | cut -d'-' -f1 | sed 's/^python//' | sort -n | tail -n1 | sed -r 's/^([0-9])([0-9]+)/\1.\2/'
+PYTHON ?= python${LATEST_PKG_PYTHON_VERSION}
```
igalic, line 8: this should still have a fallback, since we use this makefile on Travis, too.

README.md
```diff
@@ -15,11 +15,17 @@ cd libioc
...
 ```
 
-At the current time libiocage is not packaged or available in FreeBSD ports.
+The default Python version is 3.7. If you intend to run libioc from another version, please specify it during the installation:
+
+```sh
+make PYTHON=python3.8 install
+```
+
+At the current time libioc is not packaged or available in FreeBSD ports.
```
igalic, line 24: this is not entirely true…
https://www.freshports.org/devel/py-libioc it just… *still* doesn't have all the right dependencies…

Makefile
gronke: It is a bit complicated. There are multiple layers of overrides happening here.

```sh
PYTHON_VERSION ?= $(TRAVIS_PYTHON_VERSION)
SELECTED_PYTHON_VERSION != if [ "$(PYTHON_VERSION)" != "" ]; then echo $(PYTHON_VERSION); else pkg info -g 'python3*' | cut -d'-' -f1 | sed 's/^python//' | sort -n | tail -n1 | sed -r 's/^([0-9])([0-9]+)/\1.\2/'; fi
PYTHON ?= python${SELECTED_PYTHON_VERSION}
pyver= ${PYTHON:S/^python//:S/.//:C/\([0-9]+\)/\1/}
```

1.) When no `PYTHON_VERSION` was set, use `TRAVIS_PYTHON_VERSION`
2.) If neither was set, guess `SELECTED_PYTHON_VERSION` using pkg info
3.) This all can be overwritten via `PYTHON`.
4.) Read numeric `pyver` of whatever version was selected

README.md
gronke: Well, let's tackle this at a different time. Would you open an Issue as a reminder? 💜

PR #752, opened 2020-04-05
use ctypes.CDLL() directly instead of find_library

igalic:
This should be more efficient, as it doesn't need to call `ldconfig`,
`gcc`, `objdump` or `ld` to determine its existence.
`ctypes.CDDL()` will use `dlopen()` directly.

Note that we're hard-coding the current version (`libc.so.7`).
This will need changing when FreeBSD changes their ABI.

At that point we might have to rethink our strategy, or fix
`ctypes.util.find_library()` to be less awful.
Until then, this should give us a minimal performance boost.

### DIFF
--- a/libioc/Jail.py
+++ b/libioc/Jail.py
@@ -71,7 +71,11 @@ import libioc.Config.Jail.File.Fstab
 
 import ctypes
 import errno
-_dll = ctypes.CDLL("libc.so.7", use_errno=True)
+try:
+    _dll = ctypes.CDLL("libc.so.7", use_errno=True)
+except:
+    import ctypes.util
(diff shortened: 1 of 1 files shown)


libioc/Jail.py
```diff
@@ -69,9 +69,13 @@
 import libioc.ResourceSelector
 import libioc.Config.Jail.File.Fstab
 
-import ctypes.util
+import ctypes
 import errno
-_dll = ctypes.CDLL(str(ctypes.util.find_library("c")), use_errno=True)
+try:
+    _dll = ctypes.CDLL("libc.so.7", use_errno=True)
+except:
```
gronke, line 76:
```suggestion
except OSError:
```

PR #756, opened 2020-04-21
shuffle dependencies installation

igalic:
Install python3 so we'll get the default python3 installed!
That way we can now actually figure out, for real, which the default python3 is!

We also don't need `setuptools` in the standard target, so it's now moved to`install-deps-dev`.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -20,7 +20,7 @@ task:
     - sysrc -a
     - sysctl -a
     - mount -t fdescfs null /dev/fd
-    - pkg install -y git
+    - pkg install -y git-lite
     - make install-dev
   test_script:
     - env ZPOOL="ioc-test-`uname -r`" make test
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -20,7 +20,6 @@ task:
     - sysrc -a
     - sysctl -a
     - mount -t fdescfs null /dev/fd
-    - pkg install -y git-lite
     - make install-dev
   test_script:
     - env ZPOOL="ioc-test-`uname -r`" make test
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -20,6 +20,7 @@ task:
     - sysrc -a
     - sysctl -a
     - mount -t fdescfs null /dev/fd
+    - pkg install -y python3
     - make install-dev
   test_script:
     - env ZPOOL="ioc-test-`uname -r`" make test
--- a/Makefile
+++ b/Makefile
@@ -22,7 +22,7 @@ install-python-requirements:
 install-python-requirements-dev: install-python-requirements
 	$(PYTHON) -m pip install -Ur requirements-dev.txt
 install-deps:
-	pkg install -q -y libucl py$(pyver)-ucl rsync python$(pyver) py$(pyver)-libzfs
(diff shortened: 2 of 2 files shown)


Makefile
```diff
@@ -22,9 +22,9 @@ install-python-requirements:
 install-python-requirements-dev: install-python-requirements
 	$(PYTHON) -m pip install -Ur requirements-dev.txt
 install-deps:
-	pkg install -q -y libucl py$(pyver)-ucl py$(pyver)-setuptools rsync python$(pyver) py$(pyver)-libzfs
+	pkg install -q -y libucl rsync git py$(pyver)-ucl py$(pyver)-libzfs
```
gronke, line 25: I suggest we do a check for existing git executable first, and install the git-lite package if none was found.

Makefile
```diff
@@ -5,7 +5,7 @@ JAIL_NET?=16
 MYPYPATH = $(shell pwd)/.travis/mypy-stubs
 
 PYTHON_VERSION ?= $(TRAVIS_PYTHON_VERSION)
-SELECTED_PYTHON_VERSION != if [ "$(PYTHON_VERSION)" != "" ]; then echo $(PYTHON_VERSION); else pkg info -g 'python3*' | cut -d'-' -f1 | sed 's/^python//' | sort -n | tail -n1 | sed -r 's/^([0-9])([0-9]+)/\1.\2/'; fi
+SELECTED_PYTHON_VERSION != if [ "$(PYTHON_VERSION)" != "" ]; then echo $(PYTHON_VERSION); else pkg query '%dn' 'python3' | sort -un | sed -r 's/^python//;s/^([0-9])([0-9]+)/\1.\2/' | tail -n1 ; fi
```
gronke, line 8: Can you explain the expected output of `pkg query '%dn' 'python3'` please?

Here is some output from a system with python37 installed:
```console
root@freebsd-debug:~ # pkg query '%dn' 'python3'
root@freebsd-debug:~ # echo $?
69
root@freebsd-debug:~ # pkg query '%dn' 'python37'
readline
libffi
gettext-runtime
root@freebsd-debug:~ # echo $?
0
```

.cirrus.yml
```diff
@@ -20,7 +20,7 @@ task:
     - sysrc -a
     - sysctl -a
     - mount -t fdescfs null /dev/fd
-    - pkg install -y git
```
gronke, line 23: Yup, you're right: this is a global dependency, not only for Cirrus CI.

Makefile
igalic: from previous build failures, that doesn't seem to have worked.

Makefile
igalic: we explicitly `pkg install -y python3`
this pulls in the system-default python3, which should be `python37` as of 12.1-RELEASE.
*then* we check `pkg query '%dn' 'python3'` and that gives us `python37`.

Makefile
gronke:
```suggestion
	pkg install -q -y libucl rsync py$(pyver)-ucl py$(pyver)-libzfs
	if [ "`command -v git`" == "" ]; then pkg install -q -y git-list; fi 
```

gronke: Never mind. If `git-lite` is not compatible, `git` should be installed just like you did.

PR #764, opened 2020-11-13
Keep-Alive

gronke:
- refactor requirement parsing in setup.py

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -14,7 +14,8 @@ license = BSD 2-Clause License
 keywords =
     FreeBSD
     jail
-    ioc
+    iocage
+    iocell
 classifiers =
     License :: OSI Approved :: BSD 2-Clause License
     Operating System :: POSIX :: BSD :: FreeBSD
--- a/setup.cfg
+++ b/setup.cfg
@@ -32,7 +32,7 @@ packages = find:
 install_requires =
 	gitpython
 	freebsd_sysctl==0.0.7
-	jail==0.0.11
+	jail==0.0.14
 tests_require =
     pytest-runner
     pytest
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,5 +1,37 @@
 [metadata]
-description-file = README.md
+name = libioc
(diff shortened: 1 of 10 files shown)


igalic:
given that libioc is a library, we should python packaging best practices, and not have a `requirements.txt`, much less use unstable APIs to parse it

we could use the same approach i suggested in https://github.com/gronke/py-freebsd_sysctl/pull/9 and add our requirements in `setup.cfg`, and get rid of a lot of useless code

setup.cfg
```diff
@@ -1,5 +1,45 @@
...
+author = Stefan Grönke
+author_email = stefan@gronke.net
+python_requires = >=3.6
+platforms =
+    FreeBSD
+license = BSD 2-Clause License
+keywords =
+    FreeBSD
+    jail
+    ioc
```
igalic: having itself as a keyword is pointless (on some platforms it's punished) but we can add iocage and iocell as keywords here

igalic (commented):
🙋🏻‍♀️

setup.cfg
```diff
@@ -1,5 +1,45 @@
...
+    Typing :: Typed
+
+[options]
+zip_safe = False
+include_package_data = True
+packages = find:
+install_requires =
+	gitpython
+	freebsd_sysctl==0.0.7
+	jail==0.0.11
```
igalic: this keeps us only one step away from the current setup
we should use ranges in libioc, and only in ioc use exact versions of the published (on ports) packages

igalic (changes requested):
🙋🏻‍♀️

setup.cfg
gronke: I agree that major version pinning is sufficient. Meanwhile the interfaces of both tools are stable enough to not break.

igalic: time for a 1.0 release then

igalic:
why are the cirrus tests failing like that?

gronke:
superseded by #769

